PR #53, opened 2018-04-16
Cnnensemble cleanup

pdima:
This pull request contains only cleanup and changes compatible with existing models/weights.

Removed unused code, added documentation, combined level 2 models and removed low performing l2 model variants.

Removing usage of temporary files during inference, it was useful during development but not needed any more.


pjbull:
@pdima All these changes are in `re_train_models` right? If so, we can just close this PR and delete this branch now. Thanks!

PR #66, opened 2018-10-03
Add redirect for weight download, update Tensorflow version

caseyfitz:
# Main changes
1. `CACHE_DIR` Redirect for Weight Download
2. Tensorflow version upgrade to 1.10.0
3. Add `appveyor.yml` for windows build testing

## `CACHE_DIR` Redirect for Weight Download

Zamba now looks for the `CACHE_DIR` environment variable. If it's present, then zamba-specific weights are downloaded there. The `.keras` weight directory is symlinked there too. This makes it easy to use zamba in the cloud where a persistent storage drive may be mounted while the actual machine (or container) may be small. Further, this allows multiple instances of zamba to utilize the weights once they're downloaded, without having to make network calls for each instance.

## Tensorflow version upgrade to 1.10.0

This change required some casting of `Path` objects to `str`, but not much more. Import updates and some warning suppression (does not impact performance).

##  Add `appveyor.yml` for windows build testing

This change adds the `appveyor.yml` file that configures a build on AppVeyor. Doc build is not tested, only the zamba package build. Note that the xgboost binary is pulled from a github repository so this build is dependent on that link staying alive.

## Additional changes

* Some clean up of *some* paths to relay more on the config file instead of excessive `.parent` usage.
* Fix and xgboost version typo in the docs

## Note

* Once merged the `zamba-web` docker file can be changed to pull from `master`
* Reverted change to 1.11.0 may be something to keep in mind in the future, since Tf 1.11.0 adds distributed training on gpu for keras submodule using standard .fit api, which could be useful for retrain, train from scratch, and add species efforts
* Current state of this branch is what has been used to build the zamba-web docker image


zamba/models/cnnensemble/src/config.py
```diff
@@ -1,6 +1,11 @@
+from os import getenv
 from pathlib import Path
 
-MODEL_DIR = Path(__file__).parent.parent
+from dotenv import load_dotenv, find_dotenv
+
+load_dotenv(find_dotenv())
+
+MODEL_DIR = Path(__file__).parent.parent if getenv("CACHE_DIR") is None else Path(getenv("CACHE_DIR")) / "cnnensemble"
```
pjbull: In general you want your variable names in code to match the environment variable names so that it is easier to reason about.  In this case I might (1) change to `ZAMBA_CACHE_DIR` to make the env var more explicit, and then (2) have a line above that is `ZAMBA_CACHE_DIR = getenv("ZAMBA_CACHE_DIR", None)`

zamba/models/cnnensemble_model.py
```diff
@@ -131,7 +135,7 @@ def _download_weights_if_needed(self, download_region):
         # file names, paths
         fnames = ["input.tar.gz", "output.tar.gz", "data_fast.zip"]
 
-        cache_dir = Path(__file__).parent
+        cache_dir = Path(__file__).parent if getenv("CACHE_DIR") is None else getenv("CACHE_DIR")
```
pjbull: Same here as above

pjbull:
Nice! And great to see the Windows support now running.

zamba/models/cnnensemble/src/config.py
caseyfitz: ill want to make sure to change the zamba-web dockerfile too

caseyfitz:
@pjbull should be good to go now

docs/install.md
```diff
@@ -51,7 +51,7 @@ XGBoost is a library for gradient boosting trees, which is oftenused in ensemble
 
 #### XGBoost on Linux and macOS
 
-XGBoost should install with `zamba` automatically. If you see a problem with xgboost when installing zamba, the easiest fix is to run `conda install xgboost==0.7.1 -c conda-forge` in an Anaconda prompt.
+XGBoost should install with `zamba` automatically. If you see a problem with xgboost when installing zamba, the easiest fix is to run `conda install xgboost==0.71 -c conda-forge` in an Anaconda prompt.
```
pjbull: Are their other places that reference this the wrong way in the docs? I think there are. Can you grab those too?

caseyfitz:
@pjbull no `0.7.1` left now!

PR #67, opened 2018-12-13
added support for models re-training from scratch

pdima:
Added support for re-training models from scratch.

Will create the separate merge request for group aware folds split


zamba/models/cnnensemble/src/prepare_train_data.py
```diff
@@ -0,0 +1,83 @@
...
+from zamba.models.cnnensemble.src import utils
+
+
+def generate_folds(config):
+    training_set_labels_ds_full = pd.read_csv(config.TRAINING_SET_LABELS)
+    training_set_labels_ds_full['fold'] = 0
+
+    data = training_set_labels_ds_full.as_matrix(columns=config.CLASSES + ['filename', 'fold'])
+
+    cls = np.argmax(data[:, :-2], axis=1)
```
pjbull: Change variable name so this doesn't conflict with the conventional `cls` for a `Class` in `classmethod`.

zamba/models/cnnensemble/src/prepare_train_data.py
```diff
@@ -0,0 +1,83 @@
...
+
+    skf = StratifiedKFold(n_splits=len(config.TRAIN_FOLDS), shuffle=True, random_state=42)
+
+    split_idx = list(skf.split(data, cls))
+    for fold, fold_split in enumerate(split_idx):
+        items = fold_split[1]
+        data[items, -1] = fold + 1
+        print(fold)
```
pjbull: Use logging instead of print

zamba/models/cnnensemble/src/prepare_train_data.py
```diff
@@ -0,0 +1,83 @@
...
+import os
+from PIL import Image
+from multiprocessing import Pool
+from tqdm import tqdm
+
+from zamba.models.cnnensemble.src import config
+from zamba.models.cnnensemble.src import utils
+
+
+def generate_folds(config):
```
pjbull: If we pass a config, it should have a different name so it doesn't conflict with the `... import config` above.

pjbull:
Hey @pdima it looks like there are still a few of the comments from the original code review that are not yet addressed. Once those are done, this should be good to merge, thanks!

PR #70, opened 2019-07-30
Fix Appveyor CI failures

r-b-g-b:
Since the last commit (ba303ce), there have been changes to appveyor, conda, or some combination of the two that causes our build to fail with the error
```
TypeError: LoadLibrary() argument 1 must be str, not None
```

I found some discussion of the error here, https://github.com/conda/conda/issues/8836 . It doesn't give a clear answer though.

I was able to find a fix here:
https://github.com/TUW-GEO/pytesmo/pull/164/files#diff-180360612c6b8c4ed830919bbb4dd459

Although geopandas might have another solution here:
https://github.com/geopandas/geopandas/blob/master/appveyor.yml

Full disclosure: I have no idea how this fix works or how appveyor works. But I will say that this change makes the red x turn into a green dot.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,10 +10,13 @@ environment:
 
 install:
   - "set PATH=%MINICONDA%;%MINICONDA%\\Scripts;%PATH%"
+  - conda info --all
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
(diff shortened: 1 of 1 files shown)


pjbull:
Nice research! Tricksy appveyor

PR #72, opened 2019-08-08
Rewrite loading to use generator and test preds

pjbull:


zamba/models/cnnensemble/src/utils.py
```diff
@@ -119,12 +119,14 @@ def load_video_clip_frames(video_fn, frames_numbers, output_size):
     """
     X = np.zeros(shape=(len(frames_numbers),) + output_size + (3,), dtype=np.float32)
 
-    v = skvideo.io.vread(str(video_fn))
+    videogen = skvideo.io.vreader(str(video_fn))
+
     valid_frames = 0
+    for frame_ix, frame in enumerate(videogen):
+        if frame_ix > max(frames_numbers):
```
r-b-g-b: If we put

```
        if frame_ix == max(frames_numbers):
            break
```

at the end of the loop, we can read one less frame.

PR #74, opened 2019-09-05
Object recognition

r-b-g-b:
Incorporate the changes from [zamba-background-detection](https://github.com/drivendataorg/zamba-background-detection) into the main zamba package.

Major changes:
 - `zamba/convert_videos.py`: resample videos prior to processing
 - `zamba/models/mega_detector.py`: Mega object recognition that draws bounding boxes around objects in frames of a movie. Also includes code to extract features from this: number of key frames, number of detections, bounding box area
 - `zamba/models/blanknonblank.py`: The winning model from prototyping -- use the "zamba" features combined with those derived from the Mega object recognition algorithm.
 - `zamba/models/cnnensemble_model.py`: Incorporates the new blank/non-blank model to populate the "blank" probability for each video

Minor changes:
 - Instead of caching files within the package directory, use `appdirs` to get a OS-specific cache directory when `ZAMBA_CACHE_DIR` not specified
 - For the new code, use the optimized `av` based video loader. (Competition code uses the old code for now).
 - Upgraded sklearn. The zamba competition pretrained model used sklearn 0.19.1 and the new pretrained blank/non-blank model used 0.21.2. The new model will not run on 0.19.1, and the old model raises warnings about loading a pickled LabelEncoder that was generated by an older version.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,14 @@
+---
+name: Pull Request
+about: Request merging changes into the code base
+title: ''
+labels: ''
+reviewers: ''
+
+---
+
+## Description
+
+
+## Tests
+*Due to the long run times and large memory requirements, some tests cannot be completed on our CI platform. Those tests should be marked with `@pytest.skipif(zamba.config.codetree)` decorator. Run the tests locally using `pytest zamba` and paste the output below.*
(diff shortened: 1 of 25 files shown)


r-b-g-b:
@pjbull Phew, ok we're really passing now*
* codetree only**
** with new tests that don't run the whole model at once

PR #95, opened 2021-01-29
Optional resampling and blank/non-blank; updated error handling

pjbull:
## Description

Added a few parameters that significantly reduce the computation that we do by default. 

 - Adds a `--resample` flag to indicate if we should re-encode the videos to a consistent size and frame rate (default `False`).
 - Adds a `--separate_blank` flag to skip the additional blank/non-blank model that does object detection so is computationally intensive (default `False`)

We also used to try to pre-load a few frames from a video to see if it was valid. Turns out that videos can still fail to load downstream even if we are able to load the first frames. Now we catch invalid videos deeper in the algorithm code and then bookkeep those all the way back up to each ensemble member.

### DIFF
--- a/zamba/models/cnnensemble_model.py
+++ b/zamba/models/cnnensemble_model.py
@@ -70,8 +70,6 @@ class CnnEnsemble(Model):
         Returns:
             OrderedDict: A dict with `key: value` as `original path: processed path` for each video.
         """
-        self.logger.debug("Preprocessing %d videos", len(input_paths))
-
         if resample:
             self.logger.debug("Converting videos to standard resolution and frame rate.")
             output_directory = Path(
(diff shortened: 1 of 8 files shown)


zamba/models/cnnensemble_model.py
```diff
@@ -96,19 +94,41 @@ def predict(self, file_names, resample=True):
...
             )
 
+            l1_results[l1_model] = l1_model_preds
+
+            # if video is invalid for this model, track that
+            invalid_mask |= np.isnan(l1_model_preds).all(axis=(1, 2))
+
+        # remove invalid videos from l1 results:
+        l1_results = {
+            name: results[~invalid_mask, ...]
```
ejm714, line 119: what does `...` do here? is that a shortcut for selecting all other dimensions?

ejm714 (approved):
LGTM! Just one Q for my own education

zamba/models/cnnensemble_model.py
pjbull: Yep `...` gets all the other dimensions.

PR #97, opened 2021-03-17
data_path can be directory or csv with filepaths

ejm714:
## Description
Adds support for data_path pointing to a csv with a column containing filepaths that contains the videos to predict upon.

This is still WIP and is missing:
- [ ] updated documentation
- [ ] tests

## Tests
*Due to the long run times and large memory requirements, some tests cannot be completed on our CI platform. Those tests should be marked with `@pytest.skipif(zamba.config.codetree)` decorator. Run the tests locally using `pytest zamba` and paste the output below.*

### DIFF
--- a/zamba/models/cnnensemble_model.py
+++ b/zamba/models/cnnensemble_model.py
@@ -50,10 +50,16 @@ class CnnEnsemble(Model):
             self._download_weights_if_needed(download_region)
 
     def load_data(self, data_path):
-        input_paths = [
-            path for path in data_path.glob('**/*')
-            if not path.is_dir() and not path.name.startswith(".")
-        ]
+
+        if Path(data_path).is_dir():
+            input_paths = [
+                path for path in data_path.glob('**/*')
+                if not path.is_dir() and not path.name.startswith(".")
+            ]
+
+        elif Path(data_path).suffix == '.csv':
+            df = pd.read_csv(data_path)
+            input_paths = [Path(f) for f in df.filepath.values]
 
         return input_paths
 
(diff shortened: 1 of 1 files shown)


zamba/models/cnnensemble_model.py
```diff
@@ -50,10 +50,16 @@ def __init__(self,
             self._download_weights_if_needed(download_region)
 
     def load_data(self, data_path):
-        input_paths = [
-            path for path in data_path.glob('**/*')
-            if not path.is_dir() and not path.name.startswith(".")
-        ]
+
+        if Path(data_path).is_dir():
```
pjbull, line 54: Looks like we already assume `data_path` is a `Path` since it is globbed.

zamba/models/cnnensemble_model.py
```diff
@@ -50,10 +50,16 @@ def __init__(self,
...
+            input_paths = [
+                path for path in data_path.glob('**/*')
+                if not path.is_dir() and not path.name.startswith(".")
+            ]
+
+        elif Path(data_path).suffix == '.csv':
+            df = pd.read_csv(data_path)
+            input_paths = [Path(f) for f in df.filepath.values]
```
pjbull, line 62: It's pretty common for people to have newline separated list of files to work on, e.g.,:

```
vid1.mp4
vid.mp4
...
```

To handle this use case, I'm inclined to assume the _first_ column is the filenames unless the user also passes a `filepath_column` explicitly.

Also, to handle the case where there are no headers, we may want to check the first line split by "," and see if that item is a valid path. For example:

```python
with data_path.open("r") as f:
    l1 = f.readline()
    no_header = Path(l1.split(",")[0]).exists()

pd.read_csv(data_path, header=(None if no_header else 0))
```

pjbull (changes requested):

PR #100, opened 2021-04-09
Custom config with pydantic models

ejm714:
This PR lays the groundwork for supporting custom models specified via yaml config files -- both for training and predicting. There are two big chunks of work included in this PR:
- refactoring the ModelManager and using pydantic models for holding the train and predict config
- reworking the CLI to use typer instead of click and adding typed options that match the config

Tests have been added for the new CLI functionality (i.e. using yaml config files) and all existing tests have been updated to pass with the current code.

Training from custom models and predicting from custom models is not fully functional (and needs additional tests) but is tracked in 83 and 84 in zamba-algorithms. 

## Tests
*Due to the long run times and large memory requirements, some tests cannot be completed on our CI platform. Those tests should be marked with `@pytest.skipif(zamba.config.codeship)` decorator. Run the tests locally using `pytest zamba` and paste the output below.*

```
python -m pytest -s zamba
============================= test session starts ==============================
platform linux -- Python 3.6.13, pytest-5.4.3, py-1.10.0, pluggy-0.13.1
rootdir: /home/emily/zamba
plugins: cov-2.10.0, mock-3.1.1
collected 14 items

zamba/tests/test_blanknonblank.py .
zamba/tests/test_cli.py ..
Resampling videos: 100%|██████████████████████████| 1/1 [00:02<00:00,  2.52s/it]
Predicting on 1 L1 models: 100%|██████████████████| 1/1 [00:43<00:00, 43.85s/it]
Frames: 100%|█████████████████████████████████████| 1/1 [00:14<00:00, 14.43s/it]
Resampling videos: 100%|██████████████████████████| 4/4 [00:05<00:00,  1.38s/it]
Predicting on 1 L1 models: 100%|██████████████████| 1/1 [00:49<00:00, 49.84s/it]
Frames: 100%|█████████████████████████████████████| 1/1 [00:14<00:00, 14.60s/it]
Frames: 100%|█████████████████████████████████████| 1/1 [00:07<00:00,  7.62s/it]
.
Resampling videos: 100%|██████████████████████████| 1/1 [00:02<00:00,  2.29s/it]
.
Frames: 100%|█████████████████████████████████████| 2/2 [00:22<00:00, 11.41s/it]
..
zamba/tests/test_sample_model.py ...
=========== 13 passed, 1 skipped, 1819 warnings in 285.12s (0:04:45) ===========
```


zamba/models/model.py
```diff
@@ -1,19 +1,7 @@
...
-import pandas as pd
-
-
-try:
-    from tensorflow import keras
-except ImportError:
-    msg = "Zamba must have tensorflow installed, run either `pip install zamba[cpu]` "\
-          "or `pip install zamba[gpu]` " \
-          "depending on what is available on your system."
-    raise ImportError(msg)
```
pjbull, line 15: We need this check for the cnn ensemble in case someone does `pip install zamba` instead of `pip install zamba[cpu]` or `pip install zamba[gpu]`. Doesn't have to be in this file, but needs to happen on that codepath

zamba/models/model.py
```diff
@@ -47,6 +36,25 @@ def __del__(self):
         if self.delete_tempdir:
             rmtree(self.tempdir)
 
+    def load(self):
+        if self.model_path is not None:
+            if self.model_library == 'keras':
+                import keras
```
pjbull, line 42: I think that we want `from tensorflow import keras` so that we use the version that comes with tensorflow and not force that keras is installed separately

zamba/models/config.py
```diff
@@ -0,0 +1,81 @@
...
+    model_library: ModelLibraryEnum = "keras"
+    model_class: ModelClassEnum = "custom"
+    tempdir: Optional[Path] = None
+    n_epochs: Optional[int] = 10
+    height: Optional[int] = None
+    width: Optional[int] = None
+    augmentation: Optional[bool] = False
+    early_stopping: Optional[bool] = False
+    save_path: Optional[Path] = None
+    yaml: Optional[FilePath] = None
```
pjbull, line 54: What's the use case for this? It's not common to have a data structure know where it is stored. For example, if I change `n_epochs` in code, does that make sure it gets written to the `yaml` file? For that reason, I'd remove this

zamba/models/config.py
```diff
@@ -0,0 +1,81 @@
...
+    labels: FilePath = None
+    model_path: FilePath = None
+    model_library: ModelLibraryEnum = "keras"
+    model_class: ModelClassEnum = "custom"
+    tempdir: Optional[Path] = None
+    n_epochs: Optional[int] = 10
+    height: Optional[int] = None
+    width: Optional[int] = None
+    augmentation: Optional[bool] = False
+    early_stopping: Optional[bool] = False
```
pjbull, line 52: Let's pull out the unused features for now like the two above and document them in an issue. Better to track that potential work than to remove these later (which would break configs that specify these items)

zamba/models/config.py
```diff
@@ -0,0 +1,81 @@
...
+    yaml: Optional[FilePath] = None
+
+
+class FineTuneConfig(BaseModel):
+    pass
+
+
+class ModelConfig(BaseModel):
+    train_config: Optional[TrainConfig] = None
+    predict_config: Optional[PredictConfig] = None
```
pjbull, line 78: I'm a little worried about `model_path` and `model_class`, etc. clashing across these two configs and making for confusing behavior.

I think there should be a `model_config` (or `architecture_config` or `algorithm_config`) that has the common things, and then the `train_config` and `predict_config` are just the extra things you need if you are only doing that task.

zamba/models/manager.py
```diff
@@ -1,127 +1,97 @@
...
-        self.model_class = ModelName[model_class].model
-
-        self.tempdir = tempdir
-        self.model = self.model_class(model_path, verbose=verbose, **model_kwargs)
-        self.proba_threshold = proba_threshold
-
-        self.output_class_names = output_class_names
+    @staticmethod
+    def from_config(config):
+        if not isinstance(config, ModelManager):
```
pjbull, line 32: Should be `ModelConfig` not `ModelManager`

zamba/models/manager.py
```diff
@@ -1,127 +1,97 @@
...
+            config = ModelConfig.parse_file(config)
+        return ModelManager(**dict(config))
```
pjbull, line 34: We want `config.dict()` instead

zamba/tests/test_cli.py
```diff
@@ -1,33 +1,67 @@
...
 
-    result = runner.invoke(predict, [str('not/a/path'),
-                                     '--verbose'])
+    result = runner.invoke(app, ['train', '--model-library', 'fastai'])
     assert result.exit_code == 2
-    assert 'Path "not/a/path" does not exist' in result.output
+    assert "invalid choice: fastai. (choose from keras, pytorch)" in result.output
+
+    result = runner.invoke(app, ['train', '--config', str(assets_dir / 'sample_config.yaml')])
+    assert result.exit_code == 0
```
pjbull, line 67: Tests looking great 🙌

pjbull (changes requested):

zamba/models/manager.py
ejm714: hmm don't think so. IIRC this was from my learning documented here: https://drivendata.slack.com/archives/CUB01MUBT/p1617745763040200

if you pass as `config.dict()`, then you get `ModelManager(train_config={n_epochs:...}` vs if you pass `dict(config`, then you get `ModelManager(train_config=TrainConfig(n_epochs=...))` -- we want the later in order to do `self.train_config.n_epochs` in all cases rather than `self.train_config['n_epochs']` if we passed in a config file

pjbull: Hm, interesting. To me that's a signal that we're not organized quite right then. Nested Pydantic models should be serialized/deserialized at the same time. To me it's something to be wary of to depend on those classes _not_ being serialized to dicts because `dict(BaseModel)` doesn't interpret the nesting correctly.

zamba/models/config.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class ModelLibraryEnum(str, Enum):
+    keras = "keras"
+    pytorch = "pytorch"
+
+
+class ModelProfileEnum(str, Enum):
+    fast = "fast"
+    full = "full"
```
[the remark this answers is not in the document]
ejm714: They don't get a top level value in PredictConfig but are part of the model_kwargs dict default. The reason the enum is here is so we can validate and show options in the CLI https://github.com/drivendataorg/zamba/blob/custom-config/zamba/cli.py#L106

zamba/cli.py
```diff
@@ -1,164 +1,231 @@
...
+        None, help="[Not implemented] Save out trained model to this filepath."
+    ),
+    model_class: str = typer.Option(
+        "custom",
+        help="[Not implemented] Model from model zoo to train. Currently only external custom models are supported.",
+    ),
+):
+    """Train a custom model using the provided data, labels, and path to preconfigured model architecture."""
+    if config is not None:
+        manager = ModelManager.from_config(config)
```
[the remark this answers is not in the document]
ejm714: Added with `typer.echo` but wondering if we should move all the echo statements into logs

zamba/models/config.py
ejm714: dropped into https://github.com/drivendataorg/zamba-algorithms/issues/25

zamba/models/model.py
ejm714: Makes sense. I'll move it back to model.py since both SampleModel and CnnEnsemble inherit from Model (and need keras)

ejm714:
@pjbull This is ready for another look. Main changes include:
- `ModelConfig` which is separate from `TrainConfig` and `PredictConfig`
- starter `KerasModel` base class
- `from_disk` and `to_disk` methods on the KerasModel (and SampleModel)
- updated tests and argument names per changes to the configs

I was committing small chunks at a time so here's the total diff of my changes which will be easiest to review: https://github.com/drivendataorg/zamba/pull/100/files/61127548003382c8c657ad28c013d863eb2353d9..4833197dbcc8c6c495f6793560ecf3561b365b88

PR #105, opened 2021-08-20
Copy over files from zamba algorithms

ejm714:
This PR copies over the main files that will be needed for the zamba package from the zamba_algorithms repo. In addition:

- Does a preliminary pass at removing code we don't need (e.g. any models that aren't the slowfast or time dist models)
- Find and replace from zamba_algorithms to zamba to fix imports
- Moves MNIST module to `tests` since this will likely be useful in writing more model tests

Not sure how much we want to test that everything is working or if it's better to get this merged in and the fix / refine as we go forward. Thoughts?


ejm714:
Determined we need to hold off on merging until we sort out the order of operations for the development of modeling retraining code.

PR #115, opened 2021-09-29
Skip on failure, actually drop duplicates, clarify logging

ejm714:
Assortment of fixes:
- need skip_on_failure on all root validators in order to get helpful validation error and not move to the next validator which will  give us a cryptic "values[...]" is None
- actually drop duplicate filepaths so we only get one row our per prediction. currently we just say we do this but it's not implemented
- model_name has a default so if you just pass a checkpoint, model_name will be specified. we don't need to remind users of this all the time and can just tell them we're using the checkpoint

### DIFF
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -176,11 +176,9 @@ def validate_model_name_and_checkpoint(cls, values):
     if checkpoint is None and model_name is None:
         raise ValueError("Must provide either model_name or checkpoint path.")
 
-    # log a warning if user specifies both
+    # checkpoint always supercedes model since model_name cannot be None
     elif checkpoint is not None and model_name is not None:
-        logger.info(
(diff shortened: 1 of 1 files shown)


zamba/models/config.py
```diff
@@ -696,6 +694,7 @@ def validate_files(cls, values):
             logger.warning(
                 f"Found {num_duplicates} duplicate row(s) in filepaths csv. Dropping duplicates so predictions will have one row per video."
             )
+            files_df = files_df[["filepath"]].drop_duplicates()
```
pjbull, line 697: Do we want this behavior at all? What about interspecies friendships? I think we want to allow duplicates that just get coalesced during OHE since our models are multiclass anyway.

pjbull: oops, sorry. just realized this is predict not train. nevermind me

ejm714: yep this just avoids having multiple rows per video written out in preds, as i experienced

PR #117, opened 2021-09-29
Support training models with just two classes

pjbull:
Top K accuracy fails to calculate if the number of classes is 2 or fewer. In that case, we just calculate accuracy instead.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import random
+import string
 from typing import Optional, Union
 
 from loguru import logger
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -78,6 +79,21 @@ def labels_relative_path() -> os.PathLike:
     return ASSETS_DIR / "labels.csv"
 
 
+def labels_n_classes_df(n_classes):
+    """Get up a labels data frame where the labels are up to
```
ejm714, line 83:
```suggestion
    """Get up a labels dataframe where the labels are up to
```

tests/conftest.py
```diff
@@ -78,6 +79,21 @@ def labels_relative_path() -> os.PathLike:
...
+    26 classes.
+    """
+    if n_classes > len(string.ascii_uppercase):
+        raise ValueError("n_classes must be less than 26")
```
ejm714, line 87: what's the point of this?

ejm714 (approved):
LGTM, just wondering what's magic about 26

tests/conftest.py
pjbull: just cuz it's easy to implement guaranteed unique classes by just using the uppercase letters.

ejm714: ah make sense

PR #119, opened 2021-09-30
Write out splits + fix for videos that cannot be loaded

ejm714:
If you're finetuning and having splits autogenerated, you're going to want to know which files were used for training vs. holdout, particularly so you can run your trained model on the holdout/test set. 

Changes:
- writes out filepath and split column to csv in save_dir / splits.csv
- incorporates fix for videos that cannot be loaded (warn and write all zeros)

Before merge:
- [ ] add test for splits files getting written out
- [ ] add test for bad video getting zeroes

### DIFF
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -513,6 +513,13 @@ class TrainConfig(ZambaBaseModel):
                     k=len(labels),
                 )
 
+                logger.info(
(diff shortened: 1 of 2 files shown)


pjbull (approved):

pjbull:
OPened #122 for outstanding tests here

PR #132, opened 2021-10-14
Update v2 docs with new changes

klwetstone:
closes https://github.com/drivendataorg/zamba-algorithms/issues/500
closes https://github.com/drivendataorg/zamba-algorithms/issues/8

Updates docs with new changes. 

**Bonus updates**

- Incoporates changes in PR #131

`zamba/models/model_manager.py `
- Fixes bug in `predict_model`, previously had `ModelConfig(train_config=predict_config, ...)`

`zamba/models/config.py`
- Updates the docstring for `num_workers` since the default is now 3 instead of one less than the number of CPUs in the system


klwetstone:
@ejm714 feedback incorporated, ready for another look!

ejm714:
> FYI I also changed the navigation TOC to be expanded instead of bold section headers.

Pretty sure the bold section headers was an intentional change by @jayqi. Details here: https://github.com/drivendataorg/zamba/pull/129. I see rationale for both sides -- can you two strike a balance?

PR #167, opened 2021-11-05
Move yolox to object_detection module

### LINKED ISSUE #139: Resolve circular imports if models are imported in zamba/models/__init__.py
We should be able to do the following in `zamba/models/__init__.py`
```
from zamba.models.efficientnet_models import TimeDistributedEfficientNet  # noqa: F401
from zamba.models.slowfast_models import SlowFast  # noqa: F401
```

however doing so yields circular imports. For now, we import these in `zamba/__init__.py`.

See discussion here: https://github.com/drivendataorg/zamba/pull/133#discussion_r732278339

r-b-g-b:
Closes #139 

The cycle we're up against:
- data.video.load_video_frames
- models.megadetector_lite_yolox
- models.__init__
- models.efficientnet_models
- pytorch_lightning.utils.ZambaVideoClassificationLightningModule
- data.video.VideoLoaderConfig

In summary, the frame loader needs YoloX, which is in `zamba.models` so calls that `zamba.models.__init__`, which instantiates all models for the registry. For example, efficientnet loads the base class `ZambaVideoClassificationLightningModule`, which loads the `zamba.data.video.VideoLoaderConfig`. And there's our cycle.

The simplest way to resolve the circular import issue seemed to be to move YoloX code out of models and into its own `zamba.object_detection` module.

I updated the docs to reflect that; I think I caught everything.


pjbull (approved):
Code change looks great to me pending tests passing 🎉 !

ejm714:
Still curious if there's a simpler way. Seems like only importing VideoLoaderConfig in `pytorch_lightning.utils` for type checking _might_ resolve this without needing to move YOLOX?

pjbull:
I kind of like separating out object detection from the classification models, but I don't feel too strongly

docs/mkdocs.yml
```diff
@@ -46,7 +46,8 @@ nav:
           - zamba.models.model_manager: "api-reference/models-model_manager.md"
           - zamba.models.slowfast_models: "api-reference/models-slowfast_models.md"
           - zamba.models.utils: "api-reference/models-utils.md"
-          - zamba.models.yolox_models: "api-reference/models-yolox_models.md"
+      - zamba.object_detection:
```
ejm714, line 49: shouldn't `megadetector_lite_yolox.md` be in object detection as well?

r-b-g-b: I wasn't exactly sure how to arrange that. The `megadetector_lite_yolox` module does get brought into the documentation here https://deploy-preview-167--silly-keller-664934.netlify.app/api-reference/object_detection-yolox/#zamba.object_detection.yolox.megadetector_lite_yolox


But maybe adding a section would give it its own page?

ejm714: my point was that this line creates a broken link: https://github.com/drivendataorg/zamba/blob/a181700b8b814a898901f55e3f0775e7dd1b63e7/docs/mkdocs.yml#L45

ejm714:
> I kind of like separating out object detection from the classification models, but I don't feel too strongly

Agree this is a nice organization. We might end up further organizing our models directory then into classification vs. denspose vs. object detection. But this is sufficient for now

ejm714 (approved):

r-b-g-b:
@ejm714 I did try that, which does fix the `ZambaVideoClassificationLightningModule` circular dependency, but there are other circular dependencies that show up then :/ I think that points to a pretty strong cycle in the code due to the fact that `load_video_frames` requires YoloX, YoloX is in models, which through `models.__init__` ends up calling lots of model initialization code, including the dataloader, which needs `load_video_frames`. There might be another way to break the cycle by pulling the dataloader code out instead, but no guarantees that would solve it + I kind of like putting object detection in its own module.

PR #169, opened 2021-12-11
Ensure videos are allocated into all specified splits

ejm714:
This PR fixes the issue that caused the following error:

```
RuntimeError: Early stopping conditioned on metric val_macro_f1 which is not available. Pass in or modify your EarlyStopping callback to use any of the following: train_loss
```

The issue has to do with the `split` column. When we don't provide it in the labels, we use a random split. But we don't guarantee that we have all of the values from the split_proportions dict (e.g. `train` and `val` ) in the split column. With a small number of videos and the default proportions, all videos getting assigned to train (see in `splits.csv`). Since there are no val videos, there is no `val_f1_score`.

I considered implementing this without the random draw and instead calculating the number of samples that need to go into each split based on the proportions provided. However, this gets tricky to ensure based on rounding. I instead went with the while loop that tries different seeds until it finds a draw that works. If someone specifies split proportions of `train: 100, val: 1`, and only provides say 5 videos, there is risk of an infinite loop. I'm not sure how important it is to catch such a situation since 1) we don't expect people to be training on so few videos and 2) it feels hard to accidentally end up in such a situation.

**Bonus fix**: fix bug in ModelCheckpoint which lets monitor be None if there is no early stopping. `mode` must either be `min` or `max`. I set it to `min` to match PTL's default but it doesn't get used if monitor is None.

Plus cleanup to set defaults in `DummyTrainConfig` to avoid repeated code.

### DIFF
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -501,11 +501,11 @@ class TrainConfig(ZambaBaseModel):
                     "No 'site' column found so videos will be randomly allocated using split proportions."
                 )
 
-                expected_labels = [k for k, v in values["split_proportions"].items() if v > 0]
+                expected_splits = [k for k, v in values["split_proportions"].items() if v > 0]
                 labels["split"] = ""
                 seed = SPLIT_SEED
 
-                while len(np.setdiff1d(expected_labels, labels.split.unique())):
+                while len(np.setdiff1d(expected_splits, labels.split.unique())):
 
                     random.seed(seed)
                     labels["split"] = random.choices(
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -7,6 +7,7 @@ from typing import Dict, Optional, Union
 import appdirs
 import ffmpeg
 from loguru import logger
+import numpy as np
 import pandas as pd
 from pydantic import BaseModel
 from pydantic import DirectoryPath, FilePath, validator, root_validator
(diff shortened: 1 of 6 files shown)


zamba/models/config.py
```diff
@@ -495,19 +496,28 @@ def preprocess_labels(cls, values):
...
-                    k=len(labels),
+                logger.info(
+                    "No 'site' column found so videos will be randomly allocated using split proportions."
                 )
 
+                expected_labels = [k for k, v in values["split_proportions"].items() if v > 0]
+                labels["split"] = ""
+                seed = SPLIT_SEED
+
+                while len(np.setdiff1d(expected_labels, labels.split.unique())):
```
pjbull, line 508: This has the (improbable) possibility of running forever. Does it make more sense to group by the labels and then assign each video in that group randomly to a split, warning when the group is less than three items?

pjbull (changes requested):

zamba/models/config.py
```diff
@@ -495,19 +496,28 @@ def preprocess_labels(cls, values):
...
-                random.seed(SPLIT_SEED)
-                labels["split"] = random.choices(
-                    list(values["split_proportions"].keys()),
-                    weights=list(values["split_proportions"].values()),
-                    k=len(labels),
+                logger.info(
+                    "No 'site' column found so videos will be randomly allocated using split proportions."
                 )
 
+                expected_labels = [k for k, v in values["split_proportions"].items() if v > 0]
```
pjbull, line 504: This var name is a little confusing since `labels` means something specific in our context. Can we change?

ejm714: totes, will use `expected_splits`

zamba/models/config.py
pjbull: Just for clarity, I do mean the _labels_ that the videos have here (not "train", "val" and "holdout")

ejm714: this can get tricky too. say i have 5 examples of 20 different species (100 videos) and i set my split proportions to be 8/1/1. if i group by species and assign to sets based on 8/1/1 probabilities, all will go to train. however, if i do this across the whole labels file (not within species), these proportions will be valid.

ejm714: curious what your desired behavior would be given that data setup

pjbull: you could heuristically do 1 video in each to start and then random after that

ejm714: clever

ejm714:
@pjbull ready for another look. this now has the added benefit of doing split allocation within species which is something we had desired but hadn't yet implemented. this should help ensure better training results (in addition to fixing the bug)

ejm714:
@pjbull ready for reals this time!

zamba/models/config.py
```diff
@@ -495,22 +495,44 @@ def preprocess_labels(cls, values):
...
+
+                for c in labels.filter(regex="species_").columns:
+                    species_df = labels[labels[c] > 0]
+
+                    # within each species, seed splits by putting one video in each set and then allocate videos based on split proportions
+                    labels.loc[species_df.index, "split"] = expected_splits + random.choices(
+                        list(values["split_proportions"].keys()),
+                        weights=list(values["split_proportions"].values()),
+                        k=len(species_df) - len(expected_splits),
+                    )
```
pjbull, line 528: This has one weird edge case that is likely rare, so just worth filing an issue for:

```
v0.mp4, antelope
v1.mp4, antelope  # v1 assigned test by antelope grouping
v1.mp4, cow       # subsequently v1 assigned train by cow grouping
v2.mp4, antelope
v4.mp4, cow
v5.mp4, cow

# test set is now missing antelope

pjbull (approved):

PR #179, opened 2022-03-22
Separate out densepose install for contributing instructions

ejm714:
`requirements-dev.txt` no longer installs the densepose extra since DensePose requires torch to be installed first. The appropriate install and test commands for DensePose contributions are now laid out on the contributing page.

Bonus fixes
- update installation instructions to use v4 as brew will now install v5 by default
- skip densepose tests by default so that `make tests` runs with the default dev install. the verbose pytest will show that these tests are skipped

### DIFF
--- a/docs/docs/contribute/index.md
+++ b/docs/docs/contribute/index.md
@@ -20,6 +20,11 @@ $ cd zamba
 $ pip install -r requirements-dev.txt
 ```
 
+If your contribution is to the [DensePose](../models/densepose.md) model, you will need to install the additional dependencies with:
+```console
+$ pip install -e .[densepose]
+```
+
 ## Running the `zamba` test suite
 
 The included [`Makefile`](https://github.com/drivendataorg/zamba/blob/master/Makefile) contains code that uses pytest to run all tests in `zamba/tests`.
@@ -28,6 +33,11 @@ The command is (from the project root):
 
 ```console
 $ make tests
+```
+
+For [DensePose](../models/densepose.md) related tests, the command is:
+```console
+$ make densepose-tests
 ```
 
 ## Submit additional training videos
--- a/tests/test_model_manager.py
+++ b/tests/test_model_manager.py
@@ -164,7 +164,7 @@ def test_train_save_dir_overwrite(
         "train_configuration.yaml",
         "test_metrics.json",
         "val_metrics.json",
-        "epoch=0-step=10.ckpt",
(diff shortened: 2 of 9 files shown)


ejm714:
Okay @pjbull this is ready for review. Test failures are unrelated though I am looking into those

ejm714:
Update: this PR now fixes the failing the test. The issue was that some of the test assets don't have that many frames since we heavily downsampled them -- if they have fewer than 10, then selecting the 10th frame index in the single frame test will fail. I added a sort in [this commit](https://github.com/drivendataorg/zamba/pull/179/commits/a5c06e0132bf4e8e09292d75f0b479c82be9d76e) to ensure we're always using the same video for that test which has enough frames to be a valid test case.

docs/docs/contribute/index.md
```diff
@@ -30,6 +35,11 @@ The command is (from the project root):
 $ make tests
 ```
 
+For [DensePose](../models/densepose.md) related tests, the command is:
+```console
+$ make densepose-tests
```
pjbull, line 40: Should this command also set `ZAMBA_RUN_DENSEPOSE_TESTS=1` before running and include those tests? Looks like this command currently just runs the cli tests:
https://github.com/drivendataorg/zamba/blob/master/Makefile#L67-L69

pjbull (commented):
LGTM except one question about the densepose tests

docs/docs/contribute/index.md
ejm714: good catch, fixed

ejm714: and this doesn't just run the cli test but also runs the rest of the densepose tests which are contained in `tests/test_densepose.py`

tests/test_model_manager.py
```diff
@@ -164,7 +164,7 @@ def test_train_save_dir_overwrite(
         "train_configuration.yaml",
         "test_metrics.json",
         "val_metrics.json",
-        "epoch=0-step=10.ckpt",
+        "epoch=0-step=11.ckpt",
```
ejm714, line 167: Caused by recent change in 1.6.0 (released 2022-03-29)
> The filename of checkpoints created with ModelCheckpoint(filename='{step}') is different compared to previous versions. A checkpoint saved after 1 step will be named step=1.ckpt instead of step=0.ckpt (https://github.com/PyTorchLightning/pytorch-lightning/pull/11805)

PR #187, opened 2022-05-31
Use strategy instead of accelerator

ejm714:
Addresses deprecation warning:

```
LightningDeprecationWarning: Passing <pytorch_lightning.plugins.training_type.ddp.DDPPlugin object at 0x7f9b62e37cd0> `strategy` to the `plugins` flag in Trainer has been deprecated in v1.5 and will be removed in v1.7. Use `Trainer(strategy=<pytorch_lightning.plugins.training_type.ddp.DDPPlugin object at 0x7f9b62e37cd0>)` instead.
```

### DIFF
--- a/zamba/models/model_manager.py
+++ b/zamba/models/model_manager.py
@@ -268,7 +268,7 @@ def train_model(
         logger=tensorboard_logger,
         callbacks=callbacks,
         fast_dev_run=train_config.dry_run,
-        accelerator="ddp" if data_module.multiprocessing_context is not None else None,
+        strategy="ddp" if data_module.multiprocessing_context is not None else None,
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -31,6 +31,7 @@ install_requires =
     openpyxl
     pandas>1.2.0
     pandas_path
+    protobuf==3.20.1
```
ejm714, line 34: For context, see this [github action error](https://github.com/drivendataorg/zamba/runs/6680469946?check_suite_focus=true#step:7:59) and [protocol buffers update](https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates)

pjbull (approved):

PR #190, opened 2022-06-09
Fix caching

### LINKED ISSUE #189: Cached model does not get used in next run
Model caching does not prevent re-downloading of model each time. For example,

- a user specifies `model_cache_dir` in their TrainConfig, at `/my_cache`
- the model get downloaded in nested subdirectories due to default behavior of cloudpathlib
https://github.com/drivendataorg/zamba/blob/7986c417f33839c0a8d14ac66201472acbfb393a/zamba/models/model_manager.py#L84-L88

at this point, the cache directory looks like
```
└── drivendata-public-assets
    └── zamba_official_models
        └── time_distributed_7f74686b7b.ckpt
```

The next time the user runs the model, the weights will get re-downloaded because this check
https://github.com/drivendataorg/zamba/blob/7986c417f33839c0a8d14ac66201472acbfb393a/zamba/models/model_manager.py#L82
will look for the model at `/my_cache/time_distributed_7f74686b7b.ckpt`

since the checkpoint lookup is based just on the filename:

https://github.com/drivendataorg/zamba/blob/7986c417f33839c0a8d14ac66201472acbfb393a/zamba/models/utils.py#L38-L45

https://github.com/drivendataorg/zamba/blob/7986c417f33839c0a8d14ac66201472acbfb393a/zamba/models/official_models/time_distributed/config.yaml#L48

Recommended fix:
- `download_weights` should not maintain subfolders

ejm714:
Main changes:
- downloads public checkpoint file to top level model_cache_dir (this supports look ups in the correct place on future runs)
- if cached checkpoint file exists (and no other checkpoint was passed in), put the cached path on the `checkpoint` property in TrainConfig

These changes allow the previously downloaded file to actually get used on future runs. Previously, we were seeing "downloading model weights" on every run.

Closes #189

### DIFF
--- a/tests/test_densepose.py
+++ b/tests/test_densepose.py
@@ -60,7 +60,7 @@ def test_image(model, chimp_image_path, tmp_path):
         )
 
         # output to disk
-        assert anatomy_info.shape == (2, 44)
+        assert anatomy_info.shape == (1, 44)
         assert (anatomy_info > 0).any().any()
         assert (tmp_path / f"anatomized_{model}.csv").stat().st_size > 0
 
@@ -106,7 +106,7 @@ def test_video(model, chimp_video_path, tmp_path):
         )
 
         # output to disk
-        assert anatomy_info.shape == (8, 46)
+        assert anatomy_info.shape == (10, 46)
         assert (anatomy_info > 0).any().any()
         assert (tmp_path / f"anatomized_{model}.csv").stat().st_size > 0
 
--- a/tests/test_densepose.py
+++ b/tests/test_densepose.py
@@ -60,7 +60,7 @@ def test_image(model, chimp_image_path, tmp_path):
         )
 
         # output to disk
-        assert anatomy_info.shape == (1, 44)
+        assert anatomy_info.shape == (2, 44)
         assert (anatomy_info > 0).any().any()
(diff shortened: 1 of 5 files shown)


tests/test_densepose.py
```diff
@@ -60,7 +60,7 @@ def test_image(model, chimp_image_path, tmp_path):
         )
 
         # output to disk
-        assert anatomy_info.shape == (2, 44)
+        assert anatomy_info.shape == (1, 44)
```
ejm714, line 63: unrelated to caching, this is due to new densepose model release: https://github.com/facebookresearch/detectron2/tree/main/projects/DensePose#whats-new

tests/test_densepose.py
```diff
@@ -106,7 +106,7 @@ def test_video(model, chimp_video_path, tmp_path):
         )
 
         # output to disk
-        assert anatomy_info.shape == (8, 46)
+        assert anatomy_info.shape == (10, 46)
```
ejm714, line 109: unrelated to caching, this is due to new densepose model release: https://github.com/facebookresearch/detectron2/tree/main/projects/DensePose#whats-new

ejm714:
All tests are now passing, this is ready for a review

pjbull (approved):

PR #199, opened 2022-07-15
Release time distributed model trained with new frame selection

ejm714:
This PR updates the official time distributed model to one trained on the same dataset but using the new frame selection model (from https://github.com/drivendataorg/zamba/pull/195).

This was essentially a one line change of updating the directory for `time_distributed` in the `WEIGHT_LOOKUP` dictionary. All the updates to the files in the `official_models` directory happen automatically as part of the `make publish_models` command. 🎉 

Bonus fixes and adds:
- updates the s3 directories of the other two models since we shuffled some things around in the s3 bucket
- adds instructions markdown file for updating a model

Closes https://github.com/drivendataorg/pjmf-zamba/issues/96


MAINTAINING.md
```diff
@@ -0,0 +1,83 @@
+## Adding a new model
+
+TLDR; update the the `WEIGHT_LOOKUP` mapping and then run `make publish_models`.
+
+#### Weight lookup mapping
+
+The weight lookup mapping in `zamba/models/config.py` connectss each model to the private s3 directory where model training results live. For example,
```
pjbull, line 7: Typo "connectss"

pjbull: Maybe note that the "private" bucket is only used as the source of what to publish. It is not used by any of the user-facing models (all their weights come from the public buckets).

MAINTAINING.md
```diff
@@ -0,0 +1,83 @@
+## Adding a new model
```
pjbull, line 1: Maybe note that this if for adding or updating a new "official" model. Any models that users train don't need to be registered in this way to be used. That way users aren't confused by these instructions and think they need to publish their models. In fact, it may be useful for users (as opposed to "maintainers") to have a link here that points to where they can distribute their models: https://github.com/drivendataorg/zamba/wiki

pjbull (changes requested):
Typo and a couple little things for the docs!

ejm714:
Comments integrated!

pjbull (approved):

PR #202, opened 2022-08-09
201 Factor out `get_cached_array_path`

### LINKED ISSUE #201: Avoid using the entire path as part of the cache key
Relevant LOC: https://github.com/drivendataorg/zamba/blob/master/zamba/data/video.py#L359\

This line of code uses the path of the input video file as part of the path to the cached data. This path, along with the hashed configuration, effectively behaves like the cache key; that is, in order to find something in the cache, we need the same path and the same hashed configuration.

The problem is that the path contains both unique identification of the video (which is good) and also some specific information about the file system where the video is (which is bad).

For example, if the video is in an S3 bucket which is mounted on a mount point called `data`, we might see a path like this:

`s3://drivendata-client-zamba/data/raw/european/HT_T30/Left_cam47/HT_cam47_432058_5888835_20190828/08280126.MP4`

In that case, the path that's used in the cache key is

`data/raw/european/HT_T30/Left_cam47/HT_cam47_432058_5888835_20190828/08280126.MP4`

In this example, the `data/raw` part of the path should not be part of the key, because if someone else happens to mount the same bucket on a different mount point, and they look up this video, they would not find the cached data, and if they processed the video, the cached data would go in a different location.

A solution would be to strip off the part of the path prior to the geography, which we know is the beginning of the unique identifier of the video. The only sticky point is that this part of the code would need access to a complete list of geographies, which could change over time.

AllenDowney:
Factoring out `get_cached_array_path`, avoiding the pop `bug`, and adding tests.

Closes #201

### DIFF
--- a/tests/test_npy_cache.py
+++ b/tests/test_npy_cache.py
@@ -51,14 +51,14 @@ def test_get_cached_array_path():
     assert isinstance(cached_load_video_frames, type(load_video_frames))
 
     vid_path_str = "data/raw/noemie/Taï_cam197_683044_652175_20161223/01090065.AVI"
-    vid_path_path = Path(vid_path_str)
+    vid_path = Path(vid_path_str)
 
-    expected = Path(
-        "data/cache/2d1fee2b1e1f78d06aa08bdea88e7661f927bd81/data/raw/noemie/Taï_cam197_683044_652175_20161223/01090065.npy"
-    )
+    expected_cache_path = vid_path.with_suffix(".npy")
+    expected_hash = "2d1fee2b1e1f78d06aa08bdea88e7661f927bd81"
+    expected = config.cache_dir / expected_hash / expected_cache_path
 
     # test video path as string or Path
-    for video_path in [vid_path_str, vid_path_path]:
+    for video_path in [vid_path_str, vid_path]:
         path = get_cached_array_path(config, video_path)
         assert path == expected
 
@@ -66,24 +66,22 @@ def test_get_cached_array_path():
     config_dict = yaml.safe_load(config_yaml)
     config_dict["cache_dir"] = Path(config_dict["cache_dir"])
     config = VideoLoaderConfig(**config_dict)
-    path = get_cached_array_path(config, vid_path_path)
+    path = get_cached_array_path(config, vid_path)
     assert path == expected
 
     # changing config.cleanup_cache should not affect the key
     config_dict = yaml.safe_load(config_yaml)
     config_dict["cleanup_cache"] = True
     config = VideoLoaderConfig(**config_dict)
-    path = get_cached_array_path(config, vid_path_path)
+    path = get_cached_array_path(config, vid_path)
     assert path == expected
 
     # changing config.config_dir should change the path but not the hash
     config_dict = yaml.safe_load(config_yaml)
     config_dict["cache_dir"] = "something/else"
     config = VideoLoaderConfig(**config_dict)
-    path = get_cached_array_path(config, vid_path_path)
-    expected_different_path = Path(
-        "something/else/2d1fee2b1e1f78d06aa08bdea88e7661f927bd81/data/raw/noemie/Taï_cam197_683044_652175_20161223/01090065.npy"
-    )
+    path = get_cached_array_path(config, vid_path)
+    expected_different_path = config.cache_dir / expected_hash / expected_cache_path
     assert path == expected_different_path
 
     # changing anything else should change the key but not the path
@@ -91,8 +89,7 @@ def test_get_cached_array_path():
     config_dict["total_frames"] = 8
 
     config = VideoLoaderConfig(**config_dict)
-    path = get_cached_array_path(config, vid_path_path)
-    expected_different_hash = Path(
-        "data/cache/9becb6d6dfe6b9970afe05af06ef49af4881bd73/data/raw/noemie/Taï_cam197_683044_652175_20161223/01090065.npy"
-    )
+    path = get_cached_array_path(config, vid_path)
+    different_hash = "9becb6d6dfe6b9970afe05af06ef49af4881bd73"
+    expected_different_hash = config.cache_dir / different_hash / expected_cache_path
     assert path == expected_different_hash
--- a/tests/test_npy_cache.py
+++ b/tests/test_npy_cache.py
@@ -44,8 +44,8 @@ def test_get_cached_array_path():
     config_dict = yaml.safe_load(config_yaml)
     config = VideoLoaderConfig(**config_dict)
 
-    # NOTE: the validation in VideoLoaderConfig changes some fields, so this fails
-    # assert dict(config) == config_dict
+    # NOTE: the validation in VideoLoaderConfig changes some fields,
+    # so dict(config) != config_dict
 
     cached_load_video_frames = npy_cache(
         cache_path=config.cache_dir, cleanup=config.cleanup_cache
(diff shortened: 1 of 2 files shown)


tests/test_npy_cache.py
```diff
@@ -0,0 +1,112 @@
...
+    assert path == expected
+
+    # pass the cache_dir as a Path
+    path = get_cached_array_path(config, vid_path_path, Path(config.cache_dir))
+    assert path == expected
+
+    # pass the cache_dir as S3Path
+    cache_dir = S3Path("s3://drivendata-client-zamba") / config.cache_dir
+    path = get_cached_array_path(config, vid_path_path, cache_dir)
+    expected3 = S3Path(
```
ejm714, line 82: was this intended to be `expected_s3`? or just `expected`?

edit: i just saw `expected2` at the bottom. feels surprising to see expected3 before expected2

tests/test_npy_cache.py
```diff
@@ -0,0 +1,112 @@
...
+    cleanup_cache: false
+    cache_dir: data/cache
+"""
+
+
+def test_get_cached_array_path():
+    config_dict = yaml.safe_load(config_yaml)
+    config = VideoLoaderConfig(**config_dict)
+
+    # NOTE: the validation in VideoLoaderConfig changes some fields, so this fails
```
ejm714, line 47: yes this is a design issue with some of the configs, which change fields rather than setting properties based on the inputted fields. let's remove this commented out assert

tests/test_npy_cache.py
```diff
@@ -0,0 +1,112 @@
...
+    config_dict = config.dict()
+    path = get_cached_array_path(config_dict, vid_path_str, config.cache_dir)
+    assert path == expected
+
+    # pass the config as a dictionary, but use dict(config) rather than config.dict()
+    config_dict = dict(config)
+    path = get_cached_array_path(config_dict, vid_path_str, config.cache_dir)
+    # This fails because the string representation of the dictionary is different.
+    # But we don't need/want to test that it fails.
+    # assert path == expected
```
ejm714, line 97: any reason to keep this in? generally we don't keep in commented out lines of code

tests/test_npy_cache.py
```diff
@@ -0,0 +1,112 @@
...
+    path = get_cached_array_path(config_dict, vid_path_path, config.cache_dir)
+    assert path == expected
+
+    # changing anything else should
+    config_dict["total_frames"] = 8
+    path = get_cached_array_path(config_dict, vid_path_path, config.cache_dir)
+    expected2 = Path(
+        "data/cache/9becb6d6dfe6b9970afe05af06ef49af4881bd73/data/raw/noemie/Taï_cam197_683044_652175_20161223/01090065.npy"
+    )
+    assert path == expected2
```
ejm714, line 112: may be a touch clearer with

```
path_with_changed_total_frames = get_cached_array_path(config_dict, vid_path_path, config.cache_dir)
assert path_with_changed_total_frames != path
```

since the goal is to highlight that the path has changed, rather than that it matches `expected2`.

zamba/data/video.py
```diff
@@ -325,6 +325,41 @@ def validate_total_frames(cls, values):
...
+    """
+    if isinstance(config, VideoLoaderConfig):
+        # NOTE: config.dict() and dict(config) are equal, but their string representations are not
+        # so we should use config.dict() everywhere
+        config = config.dict()
+
+    # don't include `cleanup_cache` or `cache_dir` in the hashed config
+    # NOTE: sorting the keys avoids a cache miss if we see the same config in a different order
+    keys = config.keys() - {"cleanup_cache", "cache_dir"}
+    hashed_part = {k: config[k] for k in sorted(keys)}
```
ejm714, line 345: maybe a touch clearer but basically equivalent. fine to leave as is if you prefer
```
excluded_keys = ["cleanup_cache", "cache_dir"]
hashed_part = {k: config[k] for k in sorted(keys) if k not in excluded_keys}
```

zamba/data/video.py
```diff
@@ -325,6 +325,41 @@ def validate_total_frames(cls, values):
...
+    print(hashed_part)
```
ejm714, line 346: this should be logged instead of printed if we intend to keep it, but i think this is a leftover from development and can be removed

zamba/data/video.py
```diff
@@ -325,6 +325,41 @@ def validate_total_frames(cls, values):
...
+
+    config: dictionary or VideoLoaderConfig
+    vid_path: string path to the video, or Path, or S3Path
+    cache_dir: string directory name, or Path, or S3Path
+
+    returns: Path object to the cached data
+    """
+    if isinstance(config, VideoLoaderConfig):
+        # NOTE: config.dict() and dict(config) are equal, but their string representations are not
+        # so we should use config.dict() everywhere
```
ejm714, line 339: are there other places in the code where you see this? seems either worth fixing or worth filing an issue for so we remember to do it

tests/test_npy_cache.py
```diff
@@ -0,0 +1,112 @@
...
+    path = get_cached_array_path(config, vid_path_str, config.cache_dir)
+    assert path == expected
+
+    # pass the vid_path as S3Path
+    path = get_cached_array_path(config, vid_path_s3, config.cache_dir)
+    assert path == expected
+
+    # pass the vid_path as a Path
+    path = get_cached_array_path(config, vid_path_path, config.cache_dir)
+    assert path == expected
```
ejm714, line 73: curious for your thoughts on doing this as a for loop instead. the tradeoff is fewer lines of code but a little less explicit in debugging

```
# test video path as string, S3Path, and Path
for video_path in [vid_path_str, vid_path_s3, vid_path_path]:
    path = get_cached_array_path(config, video_path, config.cache_dir)
    assert path == expected
```

ejm714 (changes requested):
I really like the `get_cached_array_path` function and think the tests are good. just a couple of small comments

tests/test_npy_cache.py
AllenDowney: Half joking: that's not a commented out line of code, it's part of a note.

I agree that it's often good to remove commented out lines of code, but it is also often useful to include code snippets in documentation and notes. If I want to say, "the following line of code demonstrates the point I am making", I suppose I could put the line of code in quotes to make it clear that it is not a forbidden commented out line of code.

tests/test_npy_cache.py
AllenDowney: Same here. I spent time figuring out this gotcha, I want to warn future programmers (including me). Including the code that doesn't work seems like a natural way to do that.

zamba/data/video.py
AllenDowney: Set operations on dictionary keys are new; I think it's a good idiom to add to our repertoire.

tests/test_npy_cache.py
AllenDowney: Ok, but I also want to check the hash of a second case.

zamba/data/video.py
AllenDowney: Removed.

zamba/data/video.py
AllenDowney: The issue is that BaseModel.dict() is recursive, so sub-models are converted to dictionaries, too. 

https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeldict

That is arguably the right behavior, so I don't think we want to change it.

There are probably not many places where this is an issue. It mostly came up in testing.

But it's an argument in favor of a stricter signature for `load_video_frames`, like requiring a VideoLoaderConfig and not accepting a dictionary or kwargs.

tests/test_npy_cache.py
AllenDowney: Yes, that's better.

zamba/data/video.py
AllenDowney: I am working now on moving the cache to the right key, and I have been banging my head on this issue repeatedly. I think `get_cached_array_path` should take either a dictionary or a `VideoLoaderConfig`; supporting both is very error prone. I am leaning toward requiring `VideoLoaderConfig`. Thoughts?

ejm714: the original reason we set up `load_video_frames` to take in the config was so that we only had to document the options in one place (VideoLoaderConfig). if we support the kwargs, then we have to replicate all the documentation and make sure the options match between the two if one is added/removed.

but then we added support for a dictionary i think for ease-of-use (eg. saves a user from needing to import VideoLoaderConfig) and maybe to reduce dependency conflicts. i remember something vague about `load_video_frames` being helpful as a generic video loader, but there was some problem with because `VideoLoaderConfig` includes running Megadetector (if that option is chosen) and that requires having all the yolox dependencies. i think maybe there was something nice about being able to just specify the resizing options directly `load_video_frames(filepath, height=420, width=420)` and just use that pathway. anyway, that is all very hand-wavy because I don't remember the actual problem and so likely isn't something we should design around. 

looking through the code, i think in `zamba` we always pass in a config and don't make use of the dict (this was more when we were working in notebooks). so all that said, i'm inclined to agree with you allen

AllenDowney: Ok, I made that change and so far it is looking like a good move. Just pushed a commit with the changes.

AllenDowney:
I merged in the workaround for the problem with mkdocstrings. So I think this is ready to go now.

ejm714 (approved):
🎉

PR #205, opened 2022-08-12
Attempting to workaround an issue with mkdocstrings

AllenDowney:


AllenDowney:
@jayqi there's still one test running but it looks like this works. What do you think of this as a workaround while we wait for the dust to settle?

PR #207, opened 2022-08-16
Holding back setuptools

### LINKED ISSUE #206: New version of setuptools breaks mkdocstrings
A PR with a possible fix is #205

AllenDowney:
Closes #206

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,9 +23,9 @@ endif
 ## Install Python Dependencies
 requirements:
 ifeq (${CPU_OR_GPU}, gpu)
-	conda install cudatoolkit=11.0.3 cudnn=8.0 -c conda-forge
+	conda install -y cudatoolkit=11.0.3 cudnn=8.0 -c conda-forge
 endif
-	$(PYTHON_INTERPRETER) -m pip install -U pip setuptools wheel torch
+	$(PYTHON_INTERPRETER) -m pip install -U pip torch
 	$(PYTHON_INTERPRETER) -m pip install -r requirements-dev.txt
 
 clean: clean-build clean-pyc clean-test ## remove all build, test, coverage and Python artifacts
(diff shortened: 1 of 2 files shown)


AllenDowney:
For reasons I don't fully understand, the version of `setuptools` in the conda environment doesn't seem to matter, but the version specified in `pyproject.toml` does.

Makefile
```diff
@@ -23,9 +23,9 @@ endif
 ## Install Python Dependencies
 requirements:
 ifeq (${CPU_OR_GPU}, gpu)
-	conda install cudatoolkit=11.0.3 cudnn=8.0 -c conda-forge
+	conda install -y cudatoolkit=11.0.3 cudnn=8.0 -c conda-forge
 endif
-	$(PYTHON_INTERPRETER) -m pip install -U pip setuptools wheel torch
+	$(PYTHON_INTERPRETER) -m pip install -U pip torch
```
ejm714, line 28: have you confirmed that removing setuptools and wheel here doesn't break anything (e.g. that having it in pyproject.toml alone seems sufficient)? for example, if i pip install from the github repo and then try `import zamba`, does that work?

AllenDowney: Yes, they are already installed when the environment is created. Removing them here prevents them from being updated.

This might not be strictly necessary because the version of setuptools installed by pip appears *not* to be the version used to make the editable installation of zamba during the initial setup. However, I am reluctant to leave the installation with a version of setuptools that could break things down the road.

ejm714 (approved):
Thanks @AllenDowney! Can you file an issue to revisit unpinning setuptools if https://github.com/mkdocstrings/mkdocstrings/issues/463 gets resolved?

AllenDowney:
Created https://github.com/drivendataorg/zamba/issues/208

PR #200, opened 2022-07-20
Support binary case

ejm714:
If a user provides only two labels, we assume these are mutually exclusive and train the binary model. We log which column we're keeping.

Outstanding:
- [ ] tests
- [ ] documentation

Note: I gave this a quick try on a dataset of 100 videos balanced evenly between blank and non blank and we indeed see some learning.

```
      Validate metric             DataLoader 0
─────────────────────────────────────────────────────────
species/val_accuracy/blank            0.75
   species/val_f1/blank        0.7826086956521738
species/val_precision/blank    0.6923076923076923
 species/val_recall/blank              0.9
       val_accuracy                   0.75
         val_loss              0.6593988537788391
       val_macro_f1            0.7442455242966751
```

PR #215, opened 2022-08-22
Support binary case

AllenDowney:
Picking up where #200 left off...

If a user provides only two labels, we check that they are mutually exclusive and, if so, train the binary model. We log which column we're keeping.

Outstanding:
- [x] tests  (several of the existing tests hit the binary case)
- [ ] documentation

Note: I gave this a quick try on a dataset of 100 videos balanced evenly between blank and non blank and we indeed see some learning.

```
      Validate metric             DataLoader 0
─────────────────────────────────────────────────────────
species/val_accuracy/blank            0.75
   species/val_f1/blank        0.7826086956521738
species/val_precision/blank    0.6923076923076923
 species/val_recall/blank              0.9
       val_accuracy                   0.75
         val_loss              0.6593988537788391
       val_macro_f1            0.7442455242966751
```

Some downsides to the binary case is that these metrics can look misleading when the class are imbalanced. Not sure the best way to warn users about that. If users provide highly imbalanced data, models may learn problematically to only predict the default class, but I suppose that is the same in the multilabel case.

### DIFF
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -481,76 +481,103 @@ class TrainConfig(ZambaBaseModel):
 
     @root_validator(skip_on_failure=True)
     def preprocess_labels(cls, values):
+        """One hot encode, add splits, and check for binary case.
+
+        Replaces values['labels'] with modified DataFrame.
+
+        Args:
+            values: dictionary containing labels and other config info
+        """
         logger.info("Preprocessing labels into one hot encoded labels with one row per video.")
+        labels = values["labels"]
+
         # one hot encode collapse to one row per video
+        # TODO: replace this with a call to one_hot_encode?
         labels = (
-            pd.get_dummies(
-                values["labels"].rename(columns={"label": "species"}), columns=["species"]
-            )
+            pd.get_dummies(labels.rename(columns={"label": "species"}), columns=["species"])
             .groupby("filepath")
             .max()
         )
 
         # if no "split" column, set up train, val, and holdout split
         if "split" not in labels.columns:
-            logger.info(
-                f"Dividing videos into train, val, and holdout sets using the following split proportions: {values['split_proportions']}."
+            make_random_split(labels, values)
+
+        # if there are only two species columns and every video belongs to one of them,
+        # drop the second species column so the problem is treated as a binary classification
+        species_cols = labels.filter(regex="species_").columns
+        sums = labels[species_cols].sum(axis=1)
+
+        if len(species_cols) == 2 and (sums == 1).all():
+            logger.warning(
+                f"Binary case detected so only one species column will be kept. Output will be the binary case of {species_cols[0]}."
             )
-
-            # use site info if we have it
-            if "site" in labels.columns:
-                logger.info("Using provided 'site' column to do a site-specific split")
-                labels["split"] = create_site_specific_splits(
-                    labels["site"], proportions=values["split_proportions"]
-                )
-            else:
-                # otherwise randomly allocate
-                logger.info(
-                    "No 'site' column found so videos for each species will be randomly allocated across splits using provided split proportions."
-                )
-
-                expected_splits = [k for k, v in values["split_proportions"].items() if v > 0]
-                random.seed(SPLIT_SEED)
-
-                # check we have at least as many videos per species as we have splits
-                # labels are OHE at this point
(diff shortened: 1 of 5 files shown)


zamba/models/config.py
```diff
@@ -481,78 +481,105 @@ def validate_filepaths_and_labels(cls, values):
...
+        Replaces values['labels'] with modified DataFrame.
+
+        Args:
+            values: dictionary containing labels and other config info
+        """
         logger.info("Preprocessing labels into one hot encoded labels with one row per video.")
+        labels = values["labels"]
+
         # one hot encode collapse to one row per video
+        # TODO: replace this with a call to one_hot_encode?
```
ejm714, line 495: we could, but that [function](https://github.com/drivendataorg/pjmf-zamba/blob/main/src/data/metadata.py#L219-L247) (which lives in a separate repo so would need to be copied over) does some merging back in, assuming that there are non-label metadata columns provided. that will be the case if the user provides site or split, but if the user just provides filepath and label then that code as written will fail. i'm inclined to lave the get_dummies code as is

aside: i wonder what happens if a user provides conflicting site or split information, e.g.
```
filepath,label,site
vid_1.mp4,duiker,site_a
vid_1.mp4,bird,site_b
```
i don't think we check for that and just take the "max" for that video in this groupby (which yields one site per video). this seems like a fine assumption, though we could check this earlier and raise and error instead

zamba/models/config.py
```diff
@@ -481,78 +481,105 @@ def validate_filepaths_and_labels(cls, values):
...
+                f"Binary case detected so only one species column will be kept. Output will be the binary case of {species_cols[0]}."
+            )
+            labels = labels.drop(columns=species_cols[1])
 
         # filepath becomes column instead of index
         values["labels"] = labels.reset_index()
         return values
 
 
+def make_random_split(labels, values):
```
ejm714, line 522: not sure i'd call this `random` since it does do a site aware split if possible, otherwise it allocates videos entirely randomly. `make_split`, while more general, may be less misleading

zamba/models/config.py
```diff
@@ -481,78 +481,105 @@ def validate_filepaths_and_labels(cls, values):
...
+        logger.info(f"{labels.split.value_counts()}")
+        logger.info(f"Writing out split information to {values['save_dir'] / 'splits.csv'}.")
+
+        # create the directory to save if we need to
+        values["save_dir"].mkdir(parents=True, exist_ok=True)
+
+        labels.reset_index()[["filepath", "split"]].drop_duplicates().to_csv(
+            values["save_dir"] / "splits.csv", index=False
+        )
+
```
ejm714, line 581: this needs to return `labels`, no?

ejm714 (commented):
This is looking good, not approving yet since we'll want a test for the binary case too.

zamba/models/config.py
AllenDowney: ok

zamba/models/config.py
AllenDowney: It adds a column to labels, so I don't think it needs to return it.

But while we're here, one thing I don't understand: in the current version, the splits.csv file only gets written if we do a random split, not if we do a site-specific split. Is that correct?

I wonder if the site-specific split case is getting tested.

AllenDowney:
Looks like the failing test is

```
ImportError: Densepose not installed. See: https://zamba.drivendata.org/docs/stable/models/densepose/#installation
```

Seems like we need to either install `densepose` in the test environment or set `ZAMBA_RUN_DENSEPOSE_TESTS=0`

AllenDowney:
Looks like this was broken a few days ago: https://github.com/drivendataorg/zamba/runs/7904627719?check_suite_focus=true

zamba/models/config.py
ejm714: > But while we're here, one thing I don't understand: in the current version, the splits.csv file only gets written if we do a random split, not if we do a site-specific split. Is that correct?

Hmm good point. The intent is to only write out the splits.csv is no split is provided, but that should be both for random or site specific splits. I agree it looks like we are only doing it for the random one. Can you file an issue or fix?

AllenDowney: Fixed.

ejm714 (approved):
LGTM, @AllenDowney you can merge once the tests finish passing

PR #217, opened 2022-08-23
Hold back setuptools

AllenDowney:
The purpose of this PR is 

1) To test whether the error caused by setuptools persists, and

2) To fix with a change in `.github/workflows/tests.yml`


AllenDowney:
Apparently this is not actually necessary. Whatever was causing the problem, it is resolved now and does not require holding back setuptools in the CI env.

PR #219, opened 2022-08-24
Don't resolve filepaths

### LINKED ISSUE #218: Don't resolve filepaths
Current `data_dir` gets resolved to an absolute path, which causes caching to fail.

The fix is to leave it alone -- nothing wrong with relative paths.

AllenDowney:
Current `data_dir` gets resolved to an absolute path, which causes caching to fail.

The fix is to leave it alone -- nothing wrong with relative paths.

Closes #218

### DIFF
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -94,7 +94,6 @@ def check_files_exist_and_load(
         pd.DataFrame: DataFrame with valid and loadable videos.
     """
     # update filepath column to prepend data_dir
-    # NOTE: A previous version of this resolved all paths to absolute, but that breaks caching.
     df["filepath"] = str(data_dir) / df.filepath.path
 
     # we can have multiple rows per file with labels so limit just to one row per file for these checks
@@ -356,7 +355,7 @@ class TrainConfig(ZambaBaseModel):
     """
 
     labels: Union[FilePath, pd.DataFrame]
-    data_dir: Optional[DirectoryPath] = ""
+    data_dir: DirectoryPath = ""
     checkpoint: Optional[FilePath] = None
     scheduler_config: Optional[Union[str, SchedulerConfig]] = "default"
     model_name: Optional[ModelEnum] = ModelEnum.time_distributed
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -104,7 +104,7 @@ def check_files_exist_and_load(
     logger.info(
         f"Checking all {len(files_df):,} filepaths exist. Can take up to a minute for every couple thousand files."
     )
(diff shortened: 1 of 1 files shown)


zamba/models/config.py
```diff
@@ -104,7 +104,8 @@ def check_files_exist_and_load(
     logger.info(
         f"Checking all {len(files_df):,} filepaths exist. Can take up to a minute for every couple thousand files."
     )
-    invalid_files = files_df[~files_df.filepath.path.exists()]
+    exists = files_df['filepath'].path.exists()
```
AllenDowney, line 107: This is an unrelated change I made to help with debugging.

zamba/models/config.py
```diff
@@ -93,8 +93,8 @@ def check_files_exist_and_load(
     Returns:
         pd.DataFrame: DataFrame with valid and loadable videos.
     """
-    # update filepath column to prepend data_dir if filepath column is not an absolute path
-    data_dir = Path(data_dir).resolve()
+    # update filepath column to prepend data_dir
+    # NOTE: A previous version of this resolved all paths to absolute, but that breaks caching.
```
ejm714, line 97: [nit] do you think we need this note here? the only reason this breaks caching is because we provided a previously cached directory (which is not a supported user use case). the absolute path works fine if you let zamba handle the caching on the first epoch. i'm inclined to remove this comment since i think it contains inside baseball and doesn't help us in the future

zamba/models/config.py
```diff
@@ -355,7 +356,7 @@ class TrainConfig(ZambaBaseModel):
     """
 
     labels: Union[FilePath, pd.DataFrame]
-    data_dir: DirectoryPath = Path.cwd()
+    data_dir: Optional[DirectoryPath] = ""
```
ejm714, line 359: `Optional` means DirectoryPath can be `None` (https://pydantic-docs.helpmanual.io/usage/types/). I don't think `None` actually works otherwise you'll get `str(None) / df.filepath.path`. I think we just want this to be `DirectoryPath` (as you have in the PredictConfig)

ejm714 (changes requested):
One small tweak needed and one optional one, then it will be good to 🚀

zamba/models/config.py
AllenDowney: Agreed

zamba/models/config.py
AllenDowney: Yes.

ejm714 (approved):
LGMT! @AllenDowney feel free to merge once the tests finish passing

PR #222, opened 2022-08-28
Documentation for Windows users

pjbull:
While it is not officially supported for maintenance reasons, it is now possible to run `zamba` on a windows machine. Add the small changes that are necessary and documentation to make things work.

 - Bonus: `yolox` is now `pip install`-able, so switch to that version.


ejm714:
@pjbull can this PR be closed?

pjbull:
Temporarily fixed with #299.

#300 tracks trying to find a reliable Windows native set instructions with the right dependencies.

PR #223, opened 2022-09-06
Add ability to set learning rate

ejm714:
We should be able to specify the learning rate the same way Lightning does after it finds the desired one with [auto_lr_find](https://pytorch-lightning.readthedocs.io/en/stable/advanced/training_tricks.html#learning-rate-finder). It's simpler to do it right before training rather than in model instantiation given how complicated the `instantiate_model` function is (because we're determining if we're starting from scratch, finetuning, replacing head, etc.).

This changes one TrainConfig param. Instead of `auto_lr_find`, there is just `lr` which can either be `auto` or a float (the desired learning rate).

Needs:
- [ ] testing
- [ ] documentation

### DIFF
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -306,11 +306,11 @@ class TrainConfig(ZambaBaseModel):
             Disables tuners, checkpoint callbacks, loggers, and logger callbacks.
             Defaults to False.
         batch_size (int): Batch size to use for training. Defaults to 2.
-        auto_lr_find (bool): Use a learning rate finder algorithm when calling
-            trainer.tune() to try to find an optimal initial learning rate. Defaults to
-            False. The learning rate finder is not guaranteed to find a good learning
-            rate; depending on the dataset, it can select a learning rate that leads to
-            poor model training. Use with caution.
+        lr (bool): Learning rate for model. Defaults to 0.001. If "auto", use a
+            learning rate finder algorithm when calling trainer.tune() to try to
+            find an optimal initial learning rate. The learning rate finder is
+            not guaranteed to find a good learning rate; depending on the dataset,
+            it can select a learning rate that leads to poor model training. Use with caution.
         backbone_finetune_params (BackboneFinetuneConfig, optional): Set parameters
             to finetune a backbone model to align with the current learning rate.
             Defaults to a BackboneFinetuneConfig(unfreeze_backbone_at_epoch=5,
@@ -363,7 +363,7 @@ class TrainConfig(ZambaBaseModel):
     model_name: Optional[ModelEnum] = ModelEnum.time_distributed
     dry_run: Union[bool, int] = False
     batch_size: int = 2
-    auto_lr_find: bool = False
+    lr: Union[bool, float] = 0.001
     backbone_finetune_config: Optional[BackboneFinetuneConfig] = BackboneFinetuneConfig()
     gpus: int = GPUS_AVAILABLE
     num_workers: int = 3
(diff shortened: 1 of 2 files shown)


AllenDowney:
Adding a small fix and testing:

* With `lr: auto`, zamba tries to find the best learning rate, as requested. But you have to set `num_workers: 0` to turn off multiprocessing_context. After Epoch 0, it fails with

```
RuntimeError: Early stopping conditioned on metric `val_macro_f1` which is not available. Pass in or modify your `EarlyStopping` callback to use any of the following: `train_loss`
```

* With `lr: 0.0005` or `lr: 0.002`, we get confirmation that lr gets set:

```
2022-09-07 19:01:14.860 | INFO     | zamba.models.model_manager:train_model:285 - Setting learning rate to 0.0005.
```

But partway through Epoch 0 it gets reverted:

```
Adjusting learning rate of group 0 to 1.0000e-03.
```

AllenDowney:
@ejm714 Any thoughts?

ejm714:
> But partway through Epoch 0 it gets reverted:

Can you give more detail on when this happens? It's this at the start of epoch 0 or just randomly in the middle (which doesn't make sense)

AllenDowney:
Yes, setting `model.lr` seems to do the trick. It still adjusts `lr` during most epochs, but it seems to be always `model.lr` or `model.lr/100`

```
Epoch 0:  84%|████████████████████████████████████████████████████████▍          | 250/297 [04:20<00:48,  1.04s/it, loss=0.099, v_num=7]Adjusting learning rate of group 0 to 2.0000e-03.
Epoch 0: 100%|█████████████████████████████████████████████████████████████████████| 297/297 [04:57<00:00,  1.00s/it, loss=0.1, v_num=7Metric val_macro_f1 improved. New best score: 0.147                                                                                      
Epoch 1:  84%|████████████████████████████████████████████████████████▍          | 250/297 [09:28<01:46,  2.27s/it, loss=0.092, v_num=7]Adjusting learning rate of group 0 to 2.0000e-03.
Epoch 1: 100%|███████████████████████████████████████████████████████████████████| 297/297 [10:03<00:00,  2.03s/it, loss=0.103, v_num=7Metric val_macro_f1 improved by 0.040 >= min_delta = 0.0. New best score: 0.187                                                          
Epoch 2:  84%|███████████████████████████████████████████████████████▌          | 250/297 [14:36<02:44,  3.51s/it, loss=0.0534, v_num=7]Adjusting learning rate of group 0 to 2.0000e-03.
Epoch 2: 100%|██████████████████████████████████████████████████████████████████| 297/297 [15:13<00:00,  3.08s/it, loss=0.0485, v_num=7Metric val_macro_f1 improved by 0.001 >= min_delta = 0.0. New best score: 0.188                                                          
Epoch 2: 100%|██████████████████████████████████████████████████████████████████| 297/297 [15:13<00:00,  3.08s/it, loss=0.0485, v_num=7]Current lr: 0.002, Backbone lr: 2e-05
Epoch 3:  84%|███████████████████████████████████████████████████████▌          | 250/297 [20:43<03:53,  4.97s/it, loss=0.0688, v_num=7]Adjusting learning rate of group 0 to 2.0000e-03.
Adjusting learning rate of group 1 to 2.0000e-05.
Epoch 3: 100%|██████████████████████████████████████████████████████████████████| 297/297 [21:22<00:00,  4.32s/it, loss=0.0649, v_num=7]Current lr: 0.002, Backbone lr: 2e-05         
```

AllenDowney:
In these runs, the model reports that it is adjusting the learning rate, but the value it chooses is always the initial value that was provided; we're not seeing the legitimate-seeming adjustment we saw with `auto_lr_find: true`.

To see whether the changes we just made might be the cause of this behavior, I reverted them and ran the model again. The behavior is the same: it adjusts the learning rate near the end of each Epoch, but always chooses the initial value.

ejm714:
We need to wait until PTL fixes their bug: https://github.com/Lightning-AI/lightning/issues/14674

PR #224, opened 2022-09-07
Checking for missing files in parallel

### LINKED ISSUE #216: Speed up or eliminate checking filepaths
This step is super slow

```
Checking all 245,944 filepaths exist. Can take up to a minute for every couple thousand files.
```

Looking it utilization while it's running, `goofys` is using about 20% of a CPU, which suggests it's doing ridiculous I/O to AWS. There's gotta be a better way (or at least an option to skip it).

AllenDowney:
Closes #216 

Checking for missing files is slow with goofys. It seems to make lots of small queries to the file system. Running them in parallel with `pqdm` is much faster. The speed depends on the state of the file system cache, but we can check 246,000 files in 5-8 minutes, compared to about two hours the slow way.

Using `pqdm` with threads is faster than with processes. Using 16 threads seems to be fast and robust. With more threads, things go faster, but you start to see unpredictable I/O errors.

If an error occurs, it falls back to the slow way.

This fix has only been tested with video files that are mounted from S3 using goofys. It might be good to test with videos stored in a local file system, too.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ install_requires =
     openpyxl
     pandas>1.2.0
     pandas_path
+    pqdm
     pydantic
     python-dotenv
     pytorch-lightning>=1.6.0
--- a/zamba/models/config.py
+++ b/zamba/models/config.py
@@ -7,9 +7,11 @@ from typing import Dict, Optional, Union
 import appdirs
 import ffmpeg
 from loguru import logger
+import numpy as np
 import pandas as pd
 from pydantic import BaseModel
 from pydantic import DirectoryPath, FilePath, validator, root_validator
+from pqdm.threads import pqdm
 import torch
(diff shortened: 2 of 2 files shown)


ejm714 (commented):
Code looks great and 5-6 minutes instead of 2 hours is a fantastic improvement!

I agree it's worth testing with local files as well. Can you do that? You can just mock it by downloading one sample videos, making 100 copies or so (just a little bash script would do), and then creating a simple labels file, e.g.
```
filepath,label
vid1.mp4,gorilla
vid2.mp4,gorilla
vid3.mp4,gorilla
vid4.mp4,gorilla
```

AllenDowney:
Confirmed that it works with 22 local files (a convenience sample of videos)

ejm714 (approved):
Thanks @AllenDowney!

PR #226, opened 2022-09-13
Release time distributed model trained on ~250k videos

ejm714:
Replaces time distributed model trained on 15k videos with one trained on 250k videos. The same (improved) frame selection is used for both).

#### Key decision
This removes the 15k model so there is only one "time distributed" model available as an official model.

#### Rationale
As seen in the `val_metrics.json`, the new model does not have improved performance on every species judged by the F1 score. However, the new model has been evaluated on 30k instead of 3k videos and includes a greater diversity of videos given the inclusion of the Chimp & See data. From exploration into calibration and gains curves, it appears that there is still signal for the species with low or zero F1 score, it is just that probabilities are of lower confidence (e.g. highest probabilities still contain an animal, but those probabilities are less than 0.5). The old model was also trained on a much more balanced dataset (blanks at 15%) compared to the new model (blanks account for around 40%). The new model does significantly better on blank performance (f1 of 0.83 vs. 0.54) while also maintaining high performance on other species.

Closes https://github.com/drivendataorg/pjmf-zamba/issues/129


pjbull:
🚀 !!!

PR #228, opened 2022-09-15
Release blank / nonblank model

ejm714:
Adds a new official model: `blank_nonblank`. This was trained on both african forest and european data.

Additional changes:
- lower cases labels in provided csv to improve matching with default model labels

Outstanding:
- [x] documentation
- [x] additional validation for training this model -> see discussion in https://github.com/drivendataorg/pjmf-zamba/issues/130

Closes https://github.com/drivendataorg/pjmf-zamba/issues/130


ejm714:
@pjbull this is ready for review. You'll want to read through [this issue](https://github.com/drivendataorg/pjmf-zamba/issues/130) as well which tries to capture the implications of releasing a binary model on our other configuration options.

docs/docs/models/species-detection.md
```diff
@@ -262,6 +302,8 @@ video_loader_config:
     confidence: 0.25
     fill_mode: score_sorted
     n_frames: 32
+    image_height: 416
+    image_width: 416
```
pjbull, line 306: Update to new frame selection size (and maybe update default slowfast config?)

pjbull (changes requested):
Let's update to remove "four" and just have plurals so we don't need to change the count of models in the future!

docs/docs/models/species-detection.md
ejm714: This docs change brings the default in line with what is actually used. Whether what is used is desirable is another question.

The issue is that the official configs get generated from what was used for training, which in this case, was 416 x 416. If we change the default slowfast config manually, this will get overwritten the next time we publish models (since the official configs are generated from the full train configuration used for training, with the machine specific params removed). If we want to use slowfast with a different frame selection size, we'll need to change that workflow. 

I'm not super inclined to fix this as part of this PR which just aims to release the BNB model. It's still an open question if we continue to support the slowfast model and if we retrain it on more data, which could make this point moot.

ejm714:
@pjbull I removed the number of models count and left a comment about the slowfast model (IMO nothing to do here, if we want different behavior, let's file an issue and do that separately)

PR #229, opened 2022-09-16
Refactor branching structure for finetuning vs. resuming training

### LINKED ISSUE #212: Update the train tutorial
This page: https://zamba.drivendata.org/docs/stable/train-tutorial/

Contains some misleading/confusing information about the options, and some copy-pasted text that's actually about prediction.

ejm714:
## Refactoring in `instantiate_model`
Refactors the logic in `instantiate_model` to branch on `is_subset` to be more explicit. The previous logic tried to group the cases for replacing the head vs. resuming training. 

```
# replace the head
elif not predict_all_zamba_species or not is_subset:
...
# resume training
elif is_subset:
...
```

This new code change should not change the behavior, but should be more readable. The branching structure is now:
```
if is_subset:
  if predict_all_zamba_species:
    resume_training()
  else:
    replace_head()

else:
  replace_head()
```

## Validation for `is_subset` and `predict_all_zamba_species`
This also allows us to enforce that `is_subset` (based on the model species and the labels species) is in agreement with `predict_all_zamba_species`. Notes:
- `predict_all_zamba_species` is True by default. We assume that if you're finetuning on a subset of data, you still want all the zamba species. If you don't want this (e.g. you want duikers, elephants, and blanks only), you need to set this to False. This behavior is the same as before.
- We now only validate `predict_all_zamba_species` if you set a value (we make this field an optional boolean and use a pre validator).
- If you provide labels that are not a subset (e.g. cats and dogs) and do not specify `predict_all_zamba_labels`, we'll set predict_all_zamba_labels to False for you since your only option is a new head with a subset. This is the same result as before, where providing a subset yielded a new head. The difference is that before you could have set conflicting values, but not being a subset would have superseded the predict_all_zamba_labels field.

## Additional changes
- renames `predict_all_zamba_species` to `use_default_model_classes` to apply to the blank nonblank model as well
- fixes some copy pasta errors in the train tutorial

Outstanding:
- [x] add test for new validator

Closes #212 
Closes https://github.com/drivendataorg/pjmf-zamba/issues/130

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -528,7 +528,7 @@ def test_validate_provided_species_and_use_default_model_labels(labels_absolute_
         save_dir=tmp_path / "my_model",
         use_default_model_labels=False,
     )
-    assert config.use_default_model_labels == False
+    assert config.use_default_model_labels is False
 
     # create labels with a species that is not in the model
     alien_labels = pd.read_csv(labels_absolute_path)
@@ -541,7 +541,7 @@ def test_validate_provided_species_and_use_default_model_labels(labels_absolute_
         save_dir=tmp_path / "my_model",
     )
     # by default this gets set to False
-    assert config.use_default_model_labels == False
+    assert config.use_default_model_labels is False
 
     # if I try to set this to True, it will error
     with pytest.raises(ValueError) as error:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -528,7 +528,7 @@ def test_validate_provided_species_and_use_default_model_labels(labels_absolute_
         save_dir=tmp_path / "my_model",
         use_default_model_labels=False,
     )
-    assert config.use_default_model_labels is False
+    assert config.use_default_model_labels == False
 
     # create labels with a species that is not in the model
     alien_labels = pd.read_csv(labels_absolute_path)
@@ -541,7 +541,7 @@ def test_validate_provided_species_and_use_default_model_labels(labels_absolute_
         save_dir=tmp_path / "my_model",
     )
     # by default this gets set to False
-    assert config.use_default_model_labels is False
+    assert config.use_default_model_labels == False
 
     # if I try to set this to True, it will error
     with pytest.raises(ValueError) as error:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -507,3 +507,52 @@ def test_checkpoint_sets_model_to_default(
         skip_load_validation=True,
     )
     assert config.model_name == "dummy_model"
+
+
+def test_validate_provided_species_and_use_default_model_labels(labels_absolute_path, tmp_path):
+    # labels are subset of time distributed model
+    config = TrainConfig(
+        labels=labels_absolute_path,
+        model_name="time_distributed",
+        skip_load_validation=True,
+        save_dir=tmp_path / "my_model",
+    )
+    # because this is a subset, use_default_model_labels gets set to True
+    assert config.use_default_model_labels
+
+    # because this is a subset, I can choose to set it to False
+    config = TrainConfig(
+        labels=labels_absolute_path,
+        model_name="time_distributed",
+        skip_load_validation=True,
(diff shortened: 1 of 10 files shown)


ejm714:
This will unfortunately break a number of our tests because we use an African Forest label asset, which will trigger as "not a subset" with this validation check. We'll need to decide the best way forward here, potentially using model specific label files or mocking certain functions.

zamba/models/model_manager.py
```diff
@@ -104,60 +120,88 @@ def instantiate_model(
...
+    for c in set(hparams["species"]).difference(set(species)):
+        # labels are still OHE at this point
+        labels[f"species_{c}"] = 0
 
-        # add in remaining columns for species that are not present
-        for c in set(hparams["species"]).difference(set(species)):
-            # labels are still OHE at this point
-            labels[f"species_{c}"] = 0
+    # sort columns so columns on dataloader are the same as columns on model
+    labels.sort_index(axis=1, inplace=True)
```
pjbull, line 196: Is there a simple way to get the order from the base model and just use that rather than assuming alphabetical?

pjbull (approved):
Logic looks good to me pending tests!

ejm714:
@pjbull this required quite a bit more fiddling to get everything working again. couple changes:
- validator is regular instead of pre and the optional boolean defaults to None; we set to True if not set and labels are a subset
- further simplified instantiate_model by doing the checkpoint download when we're validating the checkpoint anyway
- first pass at standardizing how we get the default hparams for a model; basically, this looks like using the checkpoint where we have it (since we assign the public checkpoint for default models in validation), otherwise using the hparams file (only relevant where training from scratch and we explicitly don't set a checkpoint). This approach is useful in that the checkpoint consistently supersedes the base model
  - for example, you train a time distributed model with cats, dogs, and humans
  - then you go to finetune this with labels for cats and dogs; we should load the hparams from the checkpoint you specify, not the base model to check subsets and such

There are still improvements that could be made (e.g. instantiate model should probably just take the config object, we could better centralize the logic for looking up hparams), but I think this refactor is already at risk of scope creep and accomplishes the main task: ensuring that existing configuration parameters work with the new binary model.

Ready for another look

tests/test_config.py
```diff
@@ -507,3 +507,52 @@ def test_checkpoint_sets_model_to_default(
...
+
+    # because this is a subset, I can choose to set it to False
+    config = TrainConfig(
+        labels=labels_absolute_path,
+        model_name="time_distributed",
+        skip_load_validation=True,
+        save_dir=tmp_path / "my_model",
+        use_default_model_labels=False,
+    )
+    assert config.use_default_model_labels is False
```
pjbull, line 531: I think we want to assert `not` and the condition and not use `is` (or use `==`)

tests/test_config.py
```diff
@@ -507,3 +507,52 @@ def test_checkpoint_sets_model_to_default(
...
+    alien_labels["label"] = "alien"
+
+    config = TrainConfig(
+        labels=alien_labels,
+        model_name="time_distributed",
+        skip_load_validation=True,
+        save_dir=tmp_path / "my_model",
+    )
+    # by default this gets set to False
+    assert config.use_default_model_labels is False
```
pjbull, line 544: I think we want to assert `not` and the condition and not use `is` (or use `==`)

pjbull (approved):

ejm714:
It's a little hard to see from the diff on the `code review` comment but the change is basically to introduce parity in these tests: https://github.com/drivendataorg/zamba/blob/refactor-instantiation-hierarchy/tests/test_instantiate_model.py#L75-L171

The first two test explicitly if use_default_model_labels is True or False in instantiate model. And then the next two tests set up the config and pass that object to `instantiate_model` and ensure that the right thing is done using our real models

This change removes a duplicative test. Previously we had `test_finetune_new_labels` and `test_head_replaced_for_new_species` which both test head replacement. This is not a single test.

PR #245, opened 2022-11-10
Interpret empty `VIDEO_CACHE_DIR` as None

ejm714:
If `VIDEO_CACHE_DIR` gets set to an empty string or 0, it will be interpreted as True for caching. This PR adds logic so that empty strings or 0s in the environment variable will be interpreted as no caching is desired.


pjbull (approved):
LGTM!

PR #247, opened 2022-11-17
Add depth module

ejm714:
This implementation takes in a folder of videos (or csv with video filepaths) and outputs a csv of distance estimates for each frame (and detection). 

For frames with no detection, distance is null. If there are multiple detections in that frame, there is one line per detection. For example

```
filepath,time,distance
tests/assets/videos/data/raw/goualougo_2013/gorillas_2013/MPI_FID_20_Duboscia/14-Jun-2013/FID_20_Duboscia_2013-6-14_0083.AVI,0,7.4
tests/assets/videos/data/raw/goualougo_2013/gorillas_2013/MPI_FID_20_Duboscia/14-Jun-2013/FID_20_Duboscia_2013-6-14_0083.AVI,0,7.4
tests/assets/videos/data/raw/goualougo_2013/gorillas_2013/MPI_FID_20_Duboscia/14-Jun-2013/FID_20_Duboscia_2013-6-14_0083.AVI,1,
tests/assets/videos/data/raw/goualougo_2013/gorillas_2013/MPI_FID_20_Duboscia/14-Jun-2013/FID_20_Duboscia_2013-6-14_0083.AVI,2,
tests/assets/videos/data/raw/goualougo_2013/gorillas_2013/MPI_FID_20_Duboscia/14-Jun-2013/FID_20_Duboscia_2013-6-14_0083.AVI,3,
```

Not included in this PR:
- Masking out other detections where there is more than 1. Under the current implementation, if there are two detections in the frame, the distance estimate will be the same for both detections (since the same input is used for both). This allows us to capture the number of detections in the frame, but not differential distances.

Notes:
- In the original implementation, there was a bug where the order of the first two frames were swapped: [-1, -2, 0, 1, 2]. This bug is maintained to match what the model expects

Closes https://github.com/drivendataorg/pjmf-zamba/issues/57

Remaining to do:
- [x] update docs home page to include depth model (and update CLI output)

### DIFF
--- a/docs/docs/models/depth.md
+++ b/docs/docs/models/depth.md
@@ -56,7 +56,7 @@ Here's how to run the depth estimation model.
 
 ### Debugging
 
-Unlike in the species classification models, selected frames are stored in memory rather than cached to disk. If you run out of memory, try running on a smaller number of videos. If you hit a GPU memory error, try reducing the [number of workers](../../debugging/#reducing-num_workers) or the [batch size](../../debugging/#reducing-the-batch-size).
+Unlike in the species classification models, selected frames are stored in memory rather than cached to disk. If you run out of memory, try predicting on a smaller number of videos. If you hit a GPU memory error, try reducing the [number of workers](../../debugging/#reducing-num_workers) or the [batch size](../../debugging/#reducing-the-batch-size).
 
 ## Getting help
 
--- a/docs/docs/models/depth.md
+++ b/docs/docs/models/depth.md
@@ -1,0 +1,94 @@
+# Depth estimation
+
+## Background
+
+The depth estimation model is one of the winning models from the [Deep Chimpact: Depth Estimation for Wildlife Conservation](https://www.drivendata.org/competitions/82/competition-wildlife-video-depth-estimation/) machine learning challenge hosted by DrivenData. 
+
+The goal of this challenge was to use machine learning and advances in monocular (single-lens) depth estimation techniques to automatically estimate the distance between a camera trap and an animal contained in its video footage. The challenge drew on a unique labeled dataset from research teams from the Max Planck Institute for Evolutionary Anthropology (MPI-EVA) and the Wild Chimpanzee Foundation (WCF).
+
+The Zamba package supports running the depth estimation model on videos. Under the hood, it does the following:
+
+- Resamples the video to one frame per second
+- Runs the [MegadetectorLite](models/species-detection.md#megadetectorlite) model on each frame to find frames with animals in them
+- Estimates depth for each detected animal in the frame
+- Outputs a csv with predictions
+
+## Output format
+
+The output of the depth estimation model is a csv with the following columns:
+
+- `filepath`: video name
+- `time`: seconds from the start of the video
+- `distance`: distance between detected animal and the camera
+
+There will be multiple rows per timestamp if there are multiple animals detected in the frame. If there is no animal in the frame, the distance will be null.
+
+For example, the first few rows of the `depth_predictions.csv` might look like this:
+
+```
+filepath,time,distance
+video_1.avi,0,7.3504286
+video_1.avi,0,7.3504286
+video_1.avi,1,
+video_1.avi,2,
+video_1.avi,3,
+```
+
+## Installation
+
+The depth estimation is included by default. If you've already [installed zamba](/docs/install/), there's nothing more you need to do.
+
+## Running depth estimation
+
+Here's how to run the depth estimation model.
+
+=== "CLI"
+    ```bash
+    # output a csv with depth predictions for each frame in the videos in PATH_TO_VIDEOS
+    zamba depth --data-dir PATH_TO_VIDEOS
+    ```
+=== "Python"
+    ```python
+    from zamba.models.depth_estimation import DepthEstimationConfig
+    depth_conf = DepthEstimationConfig(data_dir="PATH_TO_VIDEOS")
+    depth_conf.run_model()
+    ```
+
+## Getting help
+
+To see all of the available options, run `zamba depth --help`.
+
+```console
+$ zamba depth --help
+
+ Usage: zamba depth [OPTIONS]
+
+ Run depth estimation algorithm on frames in which animals are detected.
+
+╭─ Options ─────────────────────────────────────────────────────────────────────────────────╮
+│ --filepaths                       PATH          Path to csv containing `filepath` column  │
+│                                                 with videos.                              │
+│                                                 [default: None]                           │
+│ --data-dir                        PATH          Path to folder containing videos.         │
+│                                                 [default: None]                           │
+│ --save-to                         PATH          An optional directory or csv path for     │
+│                                                 saving the output. Defaults to            │
+│                                                 `depth_predictions.csv` in the working    │
+│                                                 directory.                                │
+│                                                 [default: None]                           │
+│ --overwrite               -o                    Overwrite output csv if it exists.        │
+│ --batch-size                      INTEGER       Batch size to use for inference.          │
+│                                                 [default: None]                           │
+│ --model-cache-dir                 PATH          Path to directory for downloading model   │
+│                                                 weights. Alternatively, specify with      │
+│                                                 environment variable `MODEL_CACHE_DIR`.   │
+│                                                 If not specified, user's cache directory  │
+│                                                 is used.                                  │
+│                                                 [default: None]                           │
+│ --weight-download-region          [us|eu|asia]  Server region for downloading weights.    │
+│                                                 [default: None]                           │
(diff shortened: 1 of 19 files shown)


setup.cfg
```diff
@@ -66,7 +66,7 @@ densepose =
     detectron2-densepose @ git+https://github.com/facebookresearch/detectron2@main#subdirectory=projects/DensePose
 
 [flake8]
-ignore = E203, E501, W503
+ignore = E203, E501, W503, E722
```
ejm714, line 69: E722 allows bare excepts

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+        det_metadata = self.detection_dict[detection_idx]
+        det_frame = det_metadata["frame"]
+        det_video = det_metadata["video"]
+
+        # set up input array of frames within window of detection
+        # frames are stacked channel-wise
+        input = np.zeros((self.height, self.width, self.channels * self.num_frames))
+        n = 0
+        for frame_idx in range(det_frame - self.window_size, det_frame + self.window_size + 1):
+            input[:, :, n : n + self.channels] = self.cached_frames[det_video][
```
ejm714, line 123: worth looking into ways to be more efficient in the memory management here (e.g. no allocating ahead and doing the stack)

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                f"frame_{frame_idx}"
+            ]
+            n += self.channels
+
+        # normalize and convert to tensor
+        input = normalize(input)
```
ejm714, line 129: let's do this before we put the frame in the cached frames dict so we only have to normalize and transpose once

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+
+        Args:
+            model_cache_dir (Path): Path for downloading and saving model weights.
+            weight_download_region (str): s3 region to download pretrained weights from.
+                Options are "us" (United States), "eu" (Europe), or "asia" (Asia Pacific).
+                Defaults to "us".
+            batch_size (int, optional): Batch size to use for inference. Defaults to 64.
+                Note: a batch is a set of frames, not videos, for the depth model.
+            tta (int, optional): Number of flips to apply for test time augmentation.
+            use_log (bool, optional): Whether to take the exponential of the predictions
```
ejm714, line 155: we may want to remove this as long as we're outputting probabilities

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+        self.model_weights = model_cache_dir / model["weights"]
+        if not self.model_weights.exists():
+            model_cache_dir.mkdir(parents=True, exist_ok=True)
+            self.model_weights = download_weights(
+                model["weights"], model_cache_dir, weight_download_region
+            )
+
+        # automatically use CPU if no cuda available
+        if not torch.cuda.is_available():
+            self.device = "cpu"
```
ejm714, line 173: we should expose `device` this further up (maybe call gpus for consistency)

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+
+        # load dataset
+        test_dataset = DepthDataset(filepaths)
+        test_loader = torch.utils.data.DataLoader(
+            test_dataset,
+            batch_size=self.batch_size,
+            shuffle=False,
+            sampler=None,
+            collate_fn=None,
+            num_workers=min(self.batch_size, 8),
```
ejm714, line 191: let's expose this

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+        lengths = [np.arange(test_dataset.cached_frames[v]["video_length"]) for v in videos]
+
+        # create one row per frame for duration of video
+        output = pd.Series(index=videos, data=lengths).explode().to_frame().reset_index()
+        output.columns = ["filepath", "time"]
+
+        # merge in predictions
+        if len(predictions) > 0:
+            output = output.merge(predictions, on=["filepath", "time"], how="outer")
+        return output
```
ejm714, line 242: if no predictions (len(preds) == 0), need to add distance column (nan)

tests/test_depth.py
```diff
@@ -0,0 +1,81 @@
...
+
+    config = DepthEstimationConfig(
+        filepaths=pd.DataFrame(columns=["filepath"], data=[invalid_video]),
+        save_to=tmp_path / "preds.csv",
+    )
+    config.run_model()
+
+    # ensure outputs get written out and but is empty since video could not be loaded
+    preds = pd.read_csv(tmp_path / "preds.csv")
+    assert len(preds) == 0
```
ejm714, line 81: check that i have all three columns

tests/test_depth.py
```diff
@@ -0,0 +1,81 @@
...
+    filepaths = pd.read_csv(two_video_filepaths).filepath.values
+    preds = dem.predict(filepaths)
+
+    assert preds.shape == (77, 3)
+    assert preds.distance.notnull().sum() == 2
+    assert preds.filepath.nunique() == 2
+
+    # two detections for frame 0
+    assert preds.set_index(["filepath", "time"]).loc[filepaths[0], 0].shape[0] == 2
+
```
ejm714, line 41: add a check that the distance estimates are the same as what we get through running through the comp winning model (use a video with single animal)

ejm714 (commented):
Comments from synchronous code review

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                # if anything is detected in the frame, save out relevant frames
+                if len(detections) > 0:
+                    logger.debug(f"{len(detections)} detection(s) found at second {frame_idx}.")
+
+                    # get frame indices around frame with detection
+                    min_frame = frame_idx - self.window_size
+                    max_frame = frame_idx + self.window_size
+
+                    # add relevant resized frames to dict if not already added
+                    # if index is before start or after end of video, use an array of zeros
```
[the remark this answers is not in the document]
ejm714: comp code uses zeros: https://github.com/drivendataorg/deep-chimpact-winners/blob/fce53504040ecfa1fe5627bb2bd142ae7696a93a/2nd%20Place/test.py#L117

docs/docs/models/depth.md
```diff
@@ -0,0 +1,94 @@
+# Depth estimation
+
+## Background
```
pjbull, line 3: Can we start with a sentence like "Our depth estimation model is used to estimate the distance an animal is from the camera, which is an input into model used to estimate animal abundance."

pjbull (commented):
Docs seem great—just one thought on a clarifying intro sentence!

docs/docs/models/depth.md
ejm714: good thought, thanks

zamba/models/depth_estimation/depth_manager.py
ejm714: not doing as it causes a `[W NNPACK.cpp:53] Could not initialize NNPACK! Reason: Unsupported hardware.` error. the current implementation mimics the comp winning code so this seems reasonable enough

zamba/models/depth_estimation/depth_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                    for i in range(min_frame, max_frame + 1):
+                        if f"frame_{i}" not in cached_frames[video_filepath].keys():
+                            try:
+                                selected_frame = np.array(
+                                    # PIL expects size to be (w, h)
+                                    Image.fromarray(arr[i]).resize((self.width, self.height))
```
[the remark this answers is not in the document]
ejm714: done

ejm714:
All PR comments have been addressed. Checked memory usage with scalene and confirmed our predictions are within 0.1 of the competition code, often times closer to a difference of 0.03. This is likely because the competition code does some writing out and reading in with open cv and also uses albumentations to resize (which uses open cv under the hood). We use torchvision transforms instead. Distances are usually measured in increments of 0.5 so these differences don't seem problematic.

Tests are passing locally so I need to look into why they're failing on github.

@pjbull if you have a moment to give the diff a quick read as a backstop, that'd be great. Here's the [link](https://github.com/drivendataorg/zamba/pull/247/files/e1cef48bbdfa046067a1c28b1af935543646adcd..7bcc48afa4c44d796ca557376252cc45d20bc2ff#)

PR #262, opened 2023-03-03
Switch to pyproject.toml-based build

ejm714:
`setup.py` commands have been deprecated, so let's move to a `pyproject.toml` based build

Bonus fixes
* use the ["paths" option](https://mkdocstrings.github.io/python/usage/#using-the-paths-option) to tell mkdocstrings where to find modules instead of installing the package. This makes the test matrix run about 10 min faster overall.
* pip install pytorchvideo instead of using direct reference


jayqi:
The error in the CI failure 

> No module named build

needs

```bash
python -m pip install --upgrade pip build
```

instead of 

```bash
python -m pip install --upgrade pip setuptools wheel
```

ejm714:
> python -m pip install --upgrade pip build

We could also do `pip install -r requirements-dev.txt` as `build` is contained there and will install everything needed for the tests. Any preference @jayqi?

jayqi:
I don't have a strong preference there. Given that that part of the workflow only runs the build and doesn't run the tests, it would be faster if it didn't need to install dependencies that aren't needed for building.

ejm714:
@jayqi this is ready for your review now

pyproject.toml
```diff
@@ -1,13 +1,78 @@
...
+  "wheel",
 ]
-build-backend = "setuptools.build_meta"
+densepose = [
+  "detectron2-densepose @ git+https://github.com/facebookresearch/detectron2@main#subdirectory=projects/DensePose"
+]
+
+[project.urls]
+homepage = "https://github.com/drivendataorg/zamba"
+bug_tracker = "https://github.com/drivendataorg/zamba/issues"
```
jayqi, line 70: In case the aesthetic appearance of these links is something you care about: if you wrap these keys in double-quotes like `"Bug Tracker"` then you can have spaces in the link names.

I found out that these keys do show up as written. Example of a package that wrote out `bug-tracker` lowercase with a hyphen:

https://github.com/wntrblm/nox/blob/119fd00cbb7584127e6deae5c1e4c1c9811f850e/pyproject.toml#L56-L60

https://pypi.org/project/nox/

jayqi (approved):
Looks good. Just one note about the link names.

pyproject.toml
ejm714: thanks for the example, updated!

PR #268, opened 2023-04-11
Use concat instead of append for pandas 2.0

ejm714:
In [pandas 2.0.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html), the `.append` method got deprecated. The guidance is to use [`.concat`](https://pandas.pydata.org/docs/dev/reference/api/pandas.concat.html#pandas.concat) instead. This PR makes those updates. This only impacts our tests.


ejm714:
Failed test was just a codecov flake

Issue #198, opened 2022-07-11
Set split proportions to none if split column is provided


ejm714:
If a user just provides a split column, they'll get the default split proportions. This leads to the message
```
Labels contains split column yet split_proportions are also provided. Split column in labels takes precedence.
```
which is a bit confusing if a user doesn't intentionally provide split proportions. It'd be nice if we detect splits in the labels csv to set split proportions to None, so that this warning only shows up if a user intentionally provides both.

This won't change underlying behavior, just make the logging more intuitive.

aaronphilip19:
Hey, @sambujangfofana and I are students from the University of Michigan. We are currently working on a project wherein we have to contribute to a Github repository(https://eecs481.org/hw6.html). We are pretty interested in this issue and would want to work on it. We hope to submit a pull request this week.  Could we be assigned this issue?

slimkevo:
Could I please be assigned to this issue?

PR #269, opened 2023-04-17
attempt to fix issue 198

esterguri:
We added a check to see if the default value is being used or if the user is setting the value.


ejm714:
This PR is misunderstanding the intent of issue #198. The goal in that issue is to avoid the warning about providing split proportions if this is just the default value that is there. This PR still logs that error and just changes the default value.

The fix may entail adding a clause in the top line of this block that checks if the values are not the default ones before surfacing this error
https://github.com/drivendataorg/zamba/blob/1ebcd4168c2cc5df0df9cbb6299f7aada0f7baa4/zamba/models/config.py#L524-L529

PR #272, opened 2023-05-05
Specify map location when loading model

### LINKED ISSUE #270: Failed build on master branch (tests #578)
Workflow failed: [tests #578](https://github.com/drivendataorg/zamba/actions/runs/4841448155)

### LINKED ISSUE #271: Failed build on master branch (tests #579)
Workflow failed: [tests #579](https://github.com/drivendataorg/zamba/actions/runs/4886612895)

ejm714:
Fixes #270 
Fixes #271

### DIFF
--- a/zamba/models/model_manager.py
+++ b/zamba/models/model_manager.py
@@ -14,7 +14,6 @@ from pytorch_lightning.callbacks import EarlyStopping, ModelCheckpoint
 from pytorch_lightning.loggers import TensorBoardLogger
 from pytorch_lightning.strategies import DDPStrategy
 from pytorch_lightning.tuner import Tuner
-import torch
 
 from zamba.data.video import VideoLoaderConfig
 from zamba.models.config import (
@@ -79,8 +78,9 @@ def instantiate_model(
 
     # predicting
     if labels is None:
+        # predict; load from checkpoint uses associated hparams
         logger.info("Loading from checkpoint.")
-        model = model_class.load_from_checkpoint(checkpoint_path=checkpoint, **hparams)
+        model = model_class.load_from_checkpoint(checkpoint_path=checkpoint)
         return model
 
     # get species from labels file
@@ -110,10 +110,8 @@ def instantiate_model(
             return resume_training(
                 scheduler_config=scheduler_config,
                 hparams=hparams,
-                species=species,
                 model_class=model_class,
                 checkpoint=checkpoint,
-                labels=labels,
             )
 
         else:
@@ -157,10 +155,8 @@ def replace_head(scheduler_config, hparams, species, model_class, checkpoint):
 def resume_training(
     scheduler_config,
     hparams,
-    species,
     model_class,
     checkpoint,
-    labels,
 ):
     # resume training; add additional species columns to labels file if needed
     logger.info(
--- a/zamba/models/model_manager.py
+++ b/zamba/models/model_manager.py
@@ -42,7 +42,6 @@ def instantiate_model(
     from_scratch: Optional[bool] = None,
     model_name: Optional[ModelEnum] = None,
     use_default_model_labels: Optional[bool] = None,
-    gpus: Optional[int] = None,
 ) -> ZambaVideoClassificationLightningModule:
     """Instantiates the model from a checkpoint and detects whether the model head should be replaced.
     The model head is replaced if labels contain species that are not on the model or use_default_model_labels=False.
@@ -66,8 +65,6 @@ def instantiate_model(
             Only relevant if training from scratch.
         use_default_model_labels (bool, optional): Whether to output the full set of default model labels rather than
             just the species in the labels file. Only used if labels is not None.
-        gpus (int, optional): Number of gpus to use. If 0 is passed, cpu will be used. If None, gpus will be used
-            if detected.
 
     Returns:
         ZambaVideoClassificationLightningModule: Instantiated model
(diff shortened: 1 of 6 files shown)


pjbull:
Tests fail for me on a GPU machine with these logs:

```python
===================================================================== FAILURES ======================================================================
_________________________________________________________ test_not_use_default_model_labels _________________________________________________________

dummy_trained_model_checkpoint = PosixPath('/tmp/pytest-of-bull/pytest-0/dummy-model-dir3/my_model/version_0/dummy_model.ckpt')

    def test_not_use_default_model_labels(dummy_trained_model_checkpoint):
        """Tests that training a model using labels that are a subset of the model species but
        with use_default_model_labels=False replaces the model head."""
        original_model = DummyZambaVideoClassificationLightningModule.from_disk(
            dummy_trained_model_checkpoint
        )

        model = instantiate_model(
            checkpoint=dummy_trained_model_checkpoint,
            scheduler_config="default",
            labels=pd.DataFrame([{"filepath": "gorilla.mp4", "species_gorilla": 1}]),
            use_default_model_labels=False,
        )

>       assert (model.head.weight != original_model.head.weight).all()
E       RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

Not sure, but maybe the `from_disk` path needs to also check the device and load to the appropriate one?

zamba/models/model_manager.py
```diff
@@ -110,10 +110,8 @@ def instantiate_model(
             return resume_training(
                 scheduler_config=scheduler_config,
                 hparams=hparams,
-                species=species,
```
ejm714, line 113: [cleanup] removed unused params

ejm714:
@pjbull I learned a couple useful things in debugging this.

### 1) 
A LightningModule already keep track of whether a gpu is available. We need to specify `map_location` to satisfy torch but can use the `self.device` property rather than figuring out if there's a gpu separately. This really simplifies the code changes needed.

### 2)
`instantiate_model` has inconsistent behavior on whether the model that is returned is on the CPU or GPU. 
- If we are replacing the head or training from scratch (i.e. not just doing a plain `load_from_checkpoint`), the model is returned on CPU
- If we are just loading from a checkpoint to resume training, the newer PTL versions (e.g. 2.0.2) load the model checkpoint onto the device on which it was trained (if available). 
  - All of our pretrained models were trained on earlier versions and therefore don't have the device encoded and will by default by loaded onto the CPU. 
  - In our tests, we create a dummy model and "train" it. If the tests are run on a GPU machine, then this checkpoint will get loaded onto the GPU (again, only if we are resuming training and not replacing the head).
-  None of this functionally matters because putting the model onto the desired device is handled by `pl.Trainer` in `train_model` and `predict_model`. In addition, the purpose of the affected tests is to check the weights are correct (i.e. that the head has been correctly replaced or not), not the location of the model. Therefore, I just added a `.to("cpu")` for the single case where this discrepancy occurs.

ejm714:
I've gone ahead and made some simplifications. 

- Now, anytime we're loading from a checkpoint in the code, we use the `.from_disk` method which all models have. This always loads onto CPU. 
- Moving to GPU is handled by the `devices` parameter in `pl.Trainer`. https://github.com/drivendataorg/zamba/blob/1ebcd4168c2cc5df0df9cbb6299f7aada0f7baa4/zamba/models/model_manager.py#L282-L294
- The utility function [`configure_accelerator_and_devices_from_gpus`](https://github.com/drivendataorg/zamba/blob/1ebcd4168c2cc5df0df9cbb6299f7aada0f7baa4/zamba/models/utils.py#L80-L88) is where we ensure the user specified number of gpus gets passed to the Trainer (and the accelerator matches). In the pydantic validation earlier, we've ensured that you can't specify devices that you don't have.

Tests are passing locally in a fresh python 3.9 environment on a GPU and non-GPU machine.

zamba/pytorch_lightning/utils.py
```diff
@@ -303,5 +303,6 @@ def to_disk(self, path: os.PathLike):
         torch.save(checkpoint, path)
 
     @classmethod
-    def from_disk(cls, path: os.PathLike):
-        return cls.load_from_checkpoint(path)
+    def from_disk(cls, path: os.PathLike, **kwargs):
+        # note: we always load models onto CPU; moving to GPU is handled by `devices` in pl.Trainer
+        return cls.load_from_checkpoint(path, map_location="cpu", **kwargs)
```
pjbull, line 308: Maybe this is fine, but forcing cpu here still feels a little weird to me. What if someone isn't using the trainer? Then they have to know and move devices themselves?

Just wondering if letting things happen automatically is it a problem?

pjbull (approved):
It still seems a little weird to force cpu in all the `from_disk` calls, and seems maybe preferable to avoid that if possible. If not, the rest of the change looks great to me!

zamba/pytorch_lightning/utils.py
ejm714: > Just wondering if letting things happen automatically is it a problem?

With the recent torch changes, we need to specify a map location when loading from checkpoint (map_location cannot be None; that is the source of the failing tests). Trying to figure out if the user has a GPU and wants to use it at this point in the code just adds redundancy and complexity in a way that is not helpful (it's hard to summarize succinctly here but that is the route we initially tried and it was not a good option).

> What if someone isn't using the trainer?

Whether they're using our `train_model` code or writing their own PTL code, they'll still be using the trainer. If for some reason they're using our code to load the model but then not using PTL, then yes, they'll have to know and move devices.

PR #305, opened 2023-12-03
fixed issue 144

papapizzachess:
Added a parameter load_from_checkpoint in model_manager.py and passed the value in efficientnet_models.py


ejm714:
Thanks for taking a stab at this. For future PRs, we ask that you include tests for the changed behavior. 

I'm closing as we're in the middle of writing v3, which will most likely override this.

PR #311, opened 2024-03-01
Use uv for CI

### LINKED ISSUE #312: Failed build on master branch (tests #672)
Workflow failed: [tests #672](https://github.com/drivendataorg/zamba/actions/runs/8126335031)

ejm714:
Swaps in [drivendataorg/setup-python-uv-action](https://github.com/drivendataorg/setup-python-uv-action) and installs our dependencies with uv (without caching).

Closes #312

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,8 +71,6 @@ jobs:
         uses: drivendataorg/setup-python-uv-action@v1
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'
-          cache-dependency-path: pyproject.toml
 
       - name: Install dependencies
         run: |
@@ -86,8 +84,7 @@ jobs:
         env:
           ZAMBA_RUN_DENSEPOSE_TESTS: 1
         run: |
-          # torch is alread installed, so just add the densepose extra
-          uv pip install -e .[tests,densepose]
+          pip install -e .[densepose]
           make densepose-tests
 
       - name: Upload coverage to codecov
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -51,13 +48,13 @@ jobs:
         name: Maximize build space
         uses: easimon/maximize-build-space@master
         with:
-          root-reserve-mb: 35000  # for pip packages in /tmp
+          root-reserve-mb: 25000  # for pip packages in /tmp
```
ejm714, line 51: Looks like VM types have changed so root volume is about 10GB smaller ([source](https://github.com/easimon/maximize-build-space/issues/39#issuecomment-1935591779))

jayqi:
I think densepose fails to install because they do a build-time check of whether torch is installed, which doesn't work with modern standard of isolated builds. It looks like uv doesn't currently support installing without build isolation: https://github.com/astral-sh/uv/issues/1715

jayqi:
@ejm714 I just thought of a gotcha in how uv and setup-python-uv-action is set up. It's possible that the virtual environment created by setup-python-uv-action doesn't have pip installed in it, and so when you run `pip install densepose`, it's using the system pip. You might want to print out `which pip` and/or run `uv pip install pip` before trying to install densepose.

jayqi:
uv 0.1.16, released yesterday, now supports `--no-build-isolation` so I think you should be able to use it to install densepose. 

https://github.com/astral-sh/uv/releases/tag/0.1.16

ejm714:
In the end, looks like a 5-10 min speed up which is small in a relative sense. Tests are still the slowest part of this and we stick to pip for the installation tests (due to transitive dependencies) so overall run is still at least 30 min.

Ready for a final look @pjbull or @jayqi

pjbull (approved):
Looks good, thanks for tackling!

Issue #122, opened 2021-10-04
Add tests for splits and bad video data


pjbull:
#119 added a functionality to save out the splits if they weren't provided and to provided an all-zeros array with the correct shape if we can't load a video deep within the DataLoader. These items should have tests added for them:

- [ ] add test for splits files getting written out
- [ ] add test for bad video getting zeroes

sunny1401:
@pjbull - Can I try working on this?

ejm714:
@sunny1401 sure!

For the first one on having a `splits.csv` written out, take a look at the existing tests for other functionality having to do with splits in `test_config.py`: https://github.com/drivendataorg/zamba/blob/master/tests/test_config.py. You'll want to add a test case where splits get automatically assigned and then check that that csv gets written (per [this line](https://github.com/drivendataorg/zamba/blob/master/zamba/models/config.py#L686-L688))

papapizzachess:
I am a college student working on a software engineering project for the next weeks until the end of this year. This is my first open source project on Github. I would appreciate any information that might help.

chandrashekar0532:
Data splitting (train/validation/test) works correctly.

The model handles corrupt/invalid video files gracefully.

chandrashekar0532:
Solution Approach
1. Test Data Splitting (Train/Validation/Test)
Goal: Verify that data is correctly partitioned into splits with no leaks or overlaps.

Steps:

1. Create a Mock Dataset:

Generate a small synthetic dataset with known labels (e.g., 100 samples).

Include a split column (values: train, val, test).

2. Write tests:
def test_data_splits():
    # Load mock dataset with predefined splits
    data = pd.read_csv("mock_data.csv")
    
    # Check split proportions (e.g., 70% train, 20% val, 10% test)
    assert len(data[data["split"] == "train") / len(data) == 0.7
    assert len(data[data["split"] == "val"]) / len(data) == 0.2
    assert len(data[data["split"] == "test"]) / len(data) == 0.1
    
    # Ensure no overlap in filenames between splits
    train_files = set(data[data["split"] == "train"]["filename"])
    val_files = set(data[data["split"] == "val"]["filename"])
    assert train_files.isdisjoint(val_files)

3. Edge Cases:

Test empty splits.

Test invalid split ratios (e.g., sum ≠ 100%).

2. Test Handling of Corrupt/Invalid Video Data

Goal: Ensure the pipeline skips or gracefully handles corrupt files without crashing.

Steps:

[...]

Issue #154, opened 2021-10-25
Add hardware_dependent_fields to ZambaBaseModel's Config


pjbull:
Currently, we publish configs with our official models so that it is easy for other people to use them. We do this by removing the properties that are hardware dependent (e.g., `device: cuda` or `batch_size: 4`) and resaving the files. This happens in a function that does the work for the configs that are passed:

https://github.com/drivendataorg/zamba/blob/33c1c2a20feec695139f9c6af2ce05d67ed291a0/zamba/models/publish_models.py#L15-L49

We can simplify and generalize this code by adding a `hardware_dependent_fields` property to the config and then a `hardware_independent_dict` method to the model that will serialize the config, removing all the `hardware_dependent_fields` on that model and any children that inherit from the same base model. Here's a working implementation with a small example config:

```python
from pydantic import BaseModel


class CustomBase(BaseModel):
    class Config:
        hardware_dependent = []
        
    def dict_hardware_independent(self):
        full_dict = self.dict()
        
        # remove hardware dependent fields on children
        for field_name, field_type in self.__fields__.items():
            field_value = getattr(self, field_name, None)
            if isinstance(field_value, CustomBase):
                full_dict[field_name] = field_value.dict_hardware_independent()

        # remove hardware dependent fields on this model
        for f in self.Config.hardware_dependent:
            full_dict.pop(f)
                
        return full_dict
        

class SubModelC(CustomBase):
    x: str

    class Config:
        hardware_dependent = ['x']
        
class SubModelA(CustomBase):
    a: str
    b: str
    c: SubModelC

    class Config:
        hardware_dependent = ['a']
        
instance_a = SubModelA(a="a", b="b", c=SubModelC(x='x'))

instance_a.dict_hardware_independent()
# > {'b': 'b', 'c': {}}
```

Issue #144, opened 2021-10-21
Don't download model weights and load imagenet weights if using models for inference


pjbull:
We currently use `load_from_checkpoint` in our `ModelManager` to initialize models when doing inference. This can cause the models to download the pretrained imagenet weights from the internet even thought we don't need those. To address this, we need a parameter that we pass in to the `__init__` of the model to indicate we are doing inference/loading from a checkpoint, and then we need to pass this paramter to `load_from_checkpoint` in the `ModelManager`.

We should check across all of our models for this behavior, but this is how it works for the `time_distributed` model:

 - Here is where we need to pass a parameter indicating that we're doing inference:
https://github.com/drivendataorg/zamba/blob/7fb2a0f9599bf55bf2f538d6a0a736963cb9d9eb/zamba/models/model_manager.py#L98-L101

 - Here is where we need to use that param to skip intializing from timm:
https://github.com/drivendataorg/zamba/blob/7fb2a0f9599bf55bf2f538d6a0a736963cb9d9eb/zamba/models/efficientnet_models.py#L23-L27

ejm714:
This also arises if we are training a model that has labels which are a subset of the zamba labels, which means we "resume training" instead of replacing the head.

papapizzachess:
The code has changed a lot. Has this bug been resolved?

PR #320, opened 2024-04-26
Issues #122 and #154

peteford21:
Attempted to resolve issues #122 and #144 . Added test cases for bad video and testing splits per the comments and issue 122. Expanded on the hardware dependent configs as specified in 144.


pjbull:
Close bad PR

Issue #196, opened 2022-07-01
Confirm basic config with user before any long running validation happens


ejm714:
Right now, we check that there are at least some [filepaths are valid](https://github.com/drivendataorg/zamba/blob/master/zamba/models/config.py#L80) as part of the [TrainConfig validation](https://github.com/drivendataorg/zamba/blob/master/zamba/cli.py#L134), which happens before we [confirm](https://github.com/drivendataorg/zamba/blob/master/zamba/cli.py#L154-L181) with the user that the config is correct.

This is not ideal if you're training on a large number of mounted files (e.g. 100,000), as the filepath check can take around an hour in that case. It would be better if we could surface the config before doing any long running checks, so the user can review and adjust any needed things (e.g. add a cache directory).

aaronphilip19:
https://github.com/drivendataorg/zamba/issues/196

aaronphilip19:
Hey, @sambujangfofana and I are students from the University of Michigan. We are currently working on a project wherein we have to contribute to a Github repository(https://eecs481.org/hw6.html). We are pretty interested in this issue and would want to work on it. We hope to submit a pull request this week. Could we be assigned this issue?

PR #321, opened 2024-04-26
Surface the basic configuration with the user before proceeding with validation checking - Issue #196

sambujangfofana:
Per #196,

@aaronphilip19 and I implemented this by defining the configuration file, data directory, csv file path, and the checkpoint based on user input and default values. We then prompted the user with a message through the command line interface asking them to review the basic configuration. If they answered yes, they would proceed with the validation; if they answered no, they had the option to change whatever they needed and to run the respective command (zamba predict or zamba train) through the terminal again.

### DIFF
--- a/zamba/cli.py
+++ b/zamba/cli.py
@@ -78,6 +78,7 @@ def train(
 
     If an argument is specified in both the command line and in a yaml file, the command line input will take precedence.
     """
+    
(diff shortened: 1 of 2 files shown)


sambujangfofana:
Closed due to incorrect commits.

PR #337, opened 2024-09-05
Upgrade the required python version to python>=3.11

### LINKED ISSUE #192: Unpin thop once onnx import is resolved
In https://github.com/drivendataorg/zamba/pull/191, we pin `thop` to an earlier version since the latest causes an onnx import error. This is currently being discussed in issue https://github.com/Lyken17/pytorch-OpCounter/issues/173#issuecomment-1153194928. Once a new release is cut, we can unpin `thop`.

### LINKED ISSUE #310: Update python versions
Currently we support 3.8 and 3.9 but these are nearing [end of life](https://devguide.python.org/versions/).

We should update docs and tests to use 3.10, 3.11, and 3.12.

westford14:
PR to update the python required version to be >= 3.10 -- this required unpinning the `thop` dependency as well as changing the `ModelEnum.time_distributed` extraction of the string to `ModelEnum.time_distributed.value`.

Closes #310 
Closes #192

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.11]
+        python-version: [3.11, 3.12]
 
     steps:
       - if: matrix.os == 'ubuntu-latest'
@@ -119,7 +119,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.11]
+        python-version: [3.11, 3.12]
 
     steps:
       - name: Remove unused software
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Set up Python and uv
         uses: drivendataorg/setup-python-uv-action@v1
         with:
-          python-version: [3.11]
+          python-version: 3.11
 
       - name: Install dependencies
         run: |
@@ -119,7 +119,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.11]
+        python-version: 3.11
 
     steps:
       - name: Remove unused software
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -106,7 +106,7 @@ jobs:
           make densepose-tests
 
       - name: Upload coverage to codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           file: ./coverage.xml
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -119,7 +119,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: 3.11
+        python-version: [3.11]
 
     steps:
       - name: Remove unused software
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -101,7 +101,7 @@ jobs:
           ZAMBA_RUN_DENSEPOSE_TESTS: 1
         run: |
           uv pip install flit-core
-          # torch is alread installed, so just add the densepose extra
+          # torch is already installed, so just add the densepose extra
           uv pip install -e .[densepose] --no-build-isolation
           make densepose-tests
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Set up Python and uv
         uses: drivendataorg/setup-python-uv-action@v1
         with:
-          python-version: 3.8
+          python-version: [3.11]
 
       - name: Install dependencies
         run: |
@@ -41,7 +41,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.8, 3.9]
+        python-version: [3.11, 3.12]
 
     steps:
       - if: matrix.os == 'ubuntu-latest'
@@ -119,7 +119,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [3.8]
+        python-version: [3.11]
 
     steps:
       - name: Remove unused software
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: [3.11, 3.12]
+        python-version: [3.11]
 
     steps:
       - if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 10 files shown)


ejm714:
@westford14 this is the right direction. Let's see if we can set the floor at 3.11 instead and then use 3.11 and 3.12 in the tests. You'll need to update the version here for the tests for github actions: https://github.com/drivendataorg/zamba/blob/498bca1bd7e1cdd32f32357bab1c1657b8c86e8d/.github/workflows/tests.yml#L44

Once that's changed, I'll approve and run the github workflows for this PR

westford14:
> @westford14 this is the right direction. Let's see if we can set the floor at 3.11 instead and then use 3.11 and 3.12 in the tests. You'll need to update the version here for the tests for github actions:
> 
> https://github.com/drivendataorg/zamba/blob/498bca1bd7e1cdd32f32357bab1c1657b8c86e8d/.github/workflows/tests.yml#L44
> 
> Once that's changed, I'll approve and run the github workflows for this PR

@ejm714 -- thanks for the heads up on the github actions, i've now pinned them to 3.11 / 3.12

README.md
```diff
@@ -29,7 +29,7 @@ Visit https://zamba.drivendata.org/docs/ for full documentation and tutorials.
 
 First, make sure you have the prerequisites installed:
 
-* Python 3.8 or 3.9
+* Python >= 3.11
```
ejm714, line 32: Let's say Python 3.11 or 3.12 since we're only testing those

docs/docs/index.md
```diff
@@ -29,7 +29,7 @@ We encourage people to share their custom models trained with Zamba. If you trai
 
 First, make sure you have the prerequisites installed:
 
-* Python 3.8 or 3.9
+* Python >= 3.11
```
ejm714, line 32: Same comment as above

docs/docs/install.md
```diff
@@ -9,10 +9,10 @@ GPU configurations.
 
 Prerequisites:
 
- - Python 3.8 or 3.9
+ - Python >= 3.11
```
ejm714, line 12: Same comment as above

docs/docs/install.md
```diff
@@ -9,10 +9,10 @@ GPU configurations.
...
  - FFmpeg
 
-#### [Python](https://www.python.org/) 3.8 or 3.9
+#### [Python](https://www.python.org/) >= 3.11
```
ejm714, line 15: Same comment as above

.github/workflows/tests.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Set up Python and uv
         uses: drivendataorg/setup-python-uv-action@v1
         with:
-          python-version: 3.8
+          python-version: [3.11]
```
ejm714, line 25:
```suggestion
          python-version: 3.11
```
shouldn't be a sequence

https://github.com/drivendataorg/zamba/actions/runs/10734680787/workflow

westford14:
thanks for the comments @ejm714 -- i've updated accordingly

westford14:
@ejm714 -- sorry i fixed the matrix sequence in the github actions file

ejm714:
Thanks. That got past the initial hurdle so we actually get to run the tests on this PR however they're failing due to some installs for the densepose tests. You should be able to run the tests locally with `make test` for faster debugging. There's some details here: https://zamba.drivendata.org/docs/stable/contribute/#running-the-zamba-test-suite

westford14:
looking at the pytorch issues, only python 3.12 support is only available in pytorch >== https://github.com/pytorch/pytorch/releases/tag/v2.4.0 which breaks the densepose library dependency, i guess the best path forward here is to pin to python 3.11 

@ejm714 i've update the code accordingly

pyproject.toml
```diff
@@ -16,7 +16,7 @@ classifiers = [
   "License :: OSI Approved :: MIT License",
   "Operating System :: OS Independent",
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.11,<=3.12"
```
ejm714, line 19: looks like this is also allowing 3.12, should just be 3.11

westford14: ah sorry about that

westford14:
@ejm714 -- pinned to 3.11

westford14:
hey @ejm714 -- any changes I need to make to this PR? looks like all the tests would have passed (the `macOS 3.11` was cancelled)

westford14:
wow codecov is really ratelimiting hard here

ejm714:
> hey @ejm714 -- any changes I need to make to this PR? looks like all the tests would have passed (the macOS 3.11 was cancelled)

It's unclear. I want to make sure all tests are passing on all operating systems but as you noted, codecov is really ratelimiting and I haven't had a chance to look into why/how to fix that. If you want to speed merging along, then you could look into that. Otherwise, feel free to pull this down until I request anything specifically

westford14:
> > hey @ejm714 -- any changes I need to make to this PR? looks like all the tests would have passed (the macOS 3.11 was cancelled)
> 
> It's unclear. I want to make sure all tests are passing on all operating systems but as you noted, codecov is really ratelimiting and I haven't had a chance to look into why/how to fix that. If you want to speed merging along, then you could look into that. Otherwise, feel free to pull this down until I request anything specifically

yea i'll take a look at the ratelimit thing there -- though it does seem like the tests are passing

westford14:
@ejm714 -- was doing some research into the codecov issue and it seems to be a pretty common issue for forked repos, but in this thread someone mentioned that upgrading to the v4 github action resolved it (https://github.com/codecov/feedback/issues/358#issuecomment-2106045478) 

figured this was worth a shot before fully re-architecting this

ejm714 (changes requested):
Looks like we can actually support python 3.12

I was looking into this more and we need to remove the ceiling on `torch` in the densepose installs. The minimum torch version is 2.2.0 which is why we were getting the failures with python 3.12 and the densepose install. With the latest torch, the densepose tests are passing locally with python 3.12.

Thanks for the codecov update -- that seems to have resolved the timeouts. Once you've made the changes to support 3.11 and 3.12, I'll run the tests again and we should be good to go

westford14:
> Looks like we can actually support python 3.12
> 
> I was looking into this more and we need to remove the ceiling on `torch` in the densepose installs. The minimum torch version is 2.2.0 which is why we were getting the failures with python 3.12 and the densepose install. With the latest torch, the densepose tests are passing locally with python 3.12.
> 
> Thanks for the codecov update -- that seems to have resolved the timeouts. Once you've made the changes to support 3.11 and 3.12, I'll run the tests again and we should be good to go

awesome -- i'll do that and let you know when it's ready to rerun the tests

westford14:
@ejm714 -- just removed the cap on the densepose pytorch version, tests passed locally for me in both 3.11 and 3.12

pyproject.toml
```diff
@@ -16,7 +16,7 @@ classifiers = [
   "License :: OSI Approved :: MIT License",
   "Operating System :: OS Independent",
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.11,<3.13"
```
ejm714, line 19: we shouldn't set upper bounds here

- [guidance](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#requires-python-upper-bounds)
- [more context](https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special)

westford14: ah okay that makes sense! this is now updated @ejm714

ejm714 (approved):
Looks good, thanks!

PR #345, opened 2025-02-25
Black 2025 stable style update

jayqi:
Sets the floor on Black to 25.1.0, which has the 2025 annual stable style update.


jayqi:
Code quality passes so this should be fine.

ejm714 (approved):

Issue #278, opened 2023-07-17
Upgrade to Pydantic>=2


pjbull:
With #277 we pin Pyandtic to `<2.0.0`. This is not ideal if people want to use zamba as a library package. We've got pretty extensive and complex use of validators in the codebase, so need to ensure that everything works as expected with the upgrade.

Here is the migration guide:
 - https://docs.pydantic.dev/latest/migration/

There is an automated tool that is supposed to help do the migration for you:
 - https://github.com/pydantic/bump-pydantic

minhaj3:
Hi, I am new to open source contribution and came here by finding good first issue tag. If no one else is working on this, then could you please assign this issue to me? I have already made the changes using the bump-pydantic lib, I just need the go through to create a PR.

PR #352, opened 2025-03-17
Upgraded to Pydantic>=2

minhaj3:
Upgraded to Pydantic>=2 using bump-pydantic as mentioned in open issue #278.

There were few things which were not corrected by bump-pydantic tool, but they are not breaking changes as far as I can understand from the migration documentation.

Here are the missing/incomplete changes:
a. 3 Changes to validators: "We couldn't refactor the `validator`, please replace it by `field_validator` manually." 
b. 1 Changes to config: "The following keys were removed: `json_loads`."

Please let me know if the above missing changes are required, then I will read the migration docs in depth to understand how to correct them.

Also, since this is my very first PR, please let me know if I made any naive mistake and I would be happy to correct it.

Thanks you.

### DIFF
--- a/zamba/data/video.py
+++ b/zamba/data/video.py
@@ -16,7 +16,7 @@ import ffmpeg
 from loguru import logger
 import numpy as np
 import pandas as pd
-from pydantic import BaseModel, root_validator, validator
+from pydantic import model_validator, ConfigDict, BaseModel, validator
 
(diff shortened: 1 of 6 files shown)


pjbull:
Closing for now. If you want to work on this issue, please actually test and fix any broken functionality, don't just run the auto bump tool.

PR #361, opened 2025-03-19
Re-enable testing on `macos-latest`

### LINKED ISSUE #365: Enable OSX build testing on public repo
This is disabled in the private repo because it's expensive. Needs to reenable once it's public.

---
_Copied from [drivendataorg/zamba-wildlabs #98](https://github.com/drivendataorg/zamba-wildlabs/issues/98)_

ejm714:
The disabling in https://github.com/drivendataorg/zamba/pull/349 was a holdover from development in a private repo

Closes #365


pjbull:
Looks like the `tests` action didn't run at all because maybe there is no relevant change. We should add the workflow files to ones that trigger test runs.

ejm714:
Good call @pjbull. Looks like there are a couple mac specific errors

```
FAILED tests/test_config.py::test_predict_dry_run_and_save - FileNotFoundError: [Errno 2] No such file or directory: '/Users/runner/work/zamba/zamba/.zamba-weights/time_distributed_1d483fc723.ckpt.40BF9da1' -> '/Users/runner/work/zamba/zamba/.zamba-weights/time_distributed_1d483fc723.ckpt'
FAILED tests/test_cli.py::test_train_specific_options - AssertionError: assert 1 == 0
 +  where 1 = <Result FileNotFoundError(2, 'No such file or directory')>.exit_code
FAILED tests/test_cli.py::test_actual_prediction_on_images - AssertionError: assert 1 == 0
 +  where 1 = <Result RuntimeError('MPS backend out of memory (MPS allocated: 8.00 MB, other allocations: 16.00 KB, max allowed: 7.93 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).')>.exit_code
FAILED tests/test_images.py::test_train_integration - RuntimeError: MPS backend out of memory (MPS allocated: 8.00 MB, other allocations: 0 bytes, max allowed: 7.93 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```

ejm714:
> Nested-virtualization and Metal Performance Shaders (MPS) are not supported due to the limitation of Apple's Virtualization Framework.

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

Looks like we need to force CPU for the image tests since it's trying to use MPS and then failing

ejm714:
Test are finally passing 🎉

zamba/images/manager.py
```diff
@@ -56,7 +56,9 @@ def predict(config: ImageClassificationPredictConfig) -> None:
         ]
     )
     logger.info("Loading models")
-    detector = run_detector.load_detector("MDV5A")
+    detector = run_detector.load_detector(
+        "MDV5A", force_cpu=True if (os.getenv("RUNNER_OS") == "macOS") else False
+    )
```
pjbull, line 61: `(os.getenv("RUNNER_OS") == "macOS")`

pjbull:
Thanks for the legwork to track this mystery down!

pjbull:
Hm, looks like we take an enormous hit to CI times with this: 

image

I'm kind of on the fence about the relative value of running these on every change vs. just in a scheduled run or just on `master`.

ejm714:
Yes megadetector on CPU is very slow. I did an experiment just to run the two previously failing tests that use megadetector and those two tests alone are 7 minutes

https://github.com/drivendataorg/zamba/actions/runs/13983794993/job/39153939153

Would be very open to only running these on master or scheduled runs. Alternatively (or in addition), could also set up a workflow dispatch to run on macos so we could run on a PR if desired?

jayqi:
@ejm714 @pjbull — Per the discussion we had in Slack last week, I've updated this PR to set the strategy matrix for OS based on the GitHub Actions event type. I've also migrated from the old `setup-python-uv-action` that I had created to the official `setup-uv` action published by Astral.

pjbull:
Thanks @jayqi! Just kicked off a manually triggered macos run to test: https://github.com/drivendataorg/zamba/actions/runs/14138386980

pjbull (approved):

PR #373, opened 2025-03-19
Fix year in changelog and add missing file

ejm714:
Looks like `make docs` wasn't run here: https://github.com/drivendataorg/zamba/commit/d8677fb2d8cc34eb19b93ee774e3f9991bbeb5b0


pjbull (approved):
Thanks!

PR #378, opened 2025-04-04
CI improvements

jayqi:
- Change Python version matrix for the full test suite (slow tests) to be earliest and latest support Python versions. (Adding testing on 3.13.)
- Changes to install tests (fast tests):
    - Simplify the install tests to use `uv pip install .` instead of explicitly building and making new virtual environments unnecessarily.
    - Change Python version matrix to be full set of supported Python versions. (Adds testing on 3.13.) These tests are fast, so seems fine to do so.
    - Change OS matrix to match the OS matrix for the full test suite. 
    - Change job dependencies so that it runs in parallel with the slow tests.
- Improve the caching for model weights. (saves 2+ minutes)
    - Don't matrix on OS for model weights. We only need one set of model weights.
    - Change the behavior of this job to check for cache hit without downloading from cache. This was unnecessarily downloading model weights from the cache even though it wasn't using them for anything. This also means we can skip all of the unnecessary environment setup if there's a cache hit. 

Bonus fix: the type annotations for `PIL.Image` were wrong. `PIL.Image` is actually a module. (Pillow has bad naming conventions.)


.github/workflows/tests.yml
```diff
@@ -240,30 +262,18 @@ jobs:
...
-          $PYTHON_BIN -m zamba --help
-          $PYTHON_BIN -c "from zamba.data.video import MegadetectorLiteYoloX; MegadetectorLiteYoloX()"
+          uv pip install .
 
-      - name: Install source; test CLI + models with assets
+      - name: Test import and model loading
         run: |
-          uv venv .venv-sdist
-          PYTHON_BIN=.venv-sdist/bin/python
-          uv pip install --python=$PYTHON_BIN zamba@$(find dist -name 'zamba*.tar.gz') --force-reinstall
```
ejm714, line 260: I think we were installing the `.tar.gz` to most closely mimic what users will be installing. do you think that's unnecessary?

.github/workflows/tests.yml
```diff
@@ -240,30 +262,18 @@ jobs:
...
-          $PYTHON_BIN -m zamba --help
-          $PYTHON_BIN -c "from zamba.data.video import MegadetectorLiteYoloX; MegadetectorLiteYoloX()"
+          python -m zamba --help
+          python -c "from zamba.data.video import MegadetectorLiteYoloX; MegadetectorLiteYoloX()"
 
   notify:
     name: Notify failed build
-    needs: [code-quality, set-up-os-matrix, tests, test-install]
+    needs: [code-quality, set-up-os-matrix, test-suite, test-install]
```
ejm714, line 276: [nit] don't love this naming since in the first "test" is an adjective and in the second it's a verb

.github/workflows/tests.yml
```diff
@@ -199,17 +211,20 @@ jobs:
...
+    # Fast smoke tests
+    name: Install and loading tests (${{ matrix.os }}, Python ${{ matrix.python-version }})
+    needs: [set-up-os-matrix, cache-model-weights]
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
-        python-version: [3.11, 3.12]
+        os: ${{ fromJson(needs.set-up-os-matrix.outputs.os_matrix) }}
+        python-version: ["3.11", "3.12", "3.13"]
```
ejm714, line 223: if we're supporting python 3.13, we should add this to the README and docs

.github/workflows/tests.yml
jayqi: It _should_ be generally equivalent. In both cases, the installer (pip or uv) will install the declared build backend (in our case, `flit_core`), which will build a wheel from the source and then install the wheel. 

Having people install from sdist (the `.tar.gz` distribution) just inserts an unnecessary step of having to generate the sdist file, which is all of the same files the wheel build step is going to use but just in a different format. 

In fact, we should probably attach the wheel to the releases and have people install from the wheel file instead. It will save a step and be slightly faster for users.

.github/workflows/tests.yml
```diff
@@ -222,16 +237,23 @@ jobs:
...
           echo
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
-      - name: Setup FFmpeg
-        uses: Iamshankhadeep/setup-ffmpeg@v1.2
+      - if: matrix.os == 'ubuntu-latest'
+        name: Setup FFmpeg
+        uses: Iamshankhadeep/setup-ffmpeg@v1.1  # does not support arm64
```
jdcc, line 244: Why is arm64 support relevant here? Dunno if I'm missing some context that would be obvious to everyone with github actions experience or if this comment could be more useful, but either way I should probably know the answer.

jayqi: The `macos-latest` runner (`macos-14` and newer) are arm64 architecture. This particular action downloads pre-built binaries and the source they are using doesn't have Apple Silicon binaries. 

There might be some improvement to be made here, but this is maintaining the status quo of how our CI is currently set up.

.github/workflows/tests.yml
```diff
@@ -77,17 +77,23 @@ jobs:
...
 
     steps:
     - uses: actions/checkout@v4
 
+    - name: Check weights cache
+      id: cache-weights
+      uses: actions/cache@v4
+      with:
+        path: "${{ env.MODEL_CACHE_DIR }}"
+        key: weights-${{ hashFiles('zamba/models/official_models/**/config.yaml', 'zamba/models/densepose/densepose_manager.py') }}
```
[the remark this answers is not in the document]
jayqi: Good question. It seems like the `dl_weights` utility doesn't download the depth estimation model weights?

Not sure if this is intended or not, but maybe downloading them and adding it to the cache could be a separate PR. 

https://github.com/drivendataorg/zamba/blob/master/zamba/utils_cli.py

pjbull (approved):
No additional comments from me, thanks!

Issue #180, opened 2022-03-23
Support ffmpeg 5


ejm714:
Brew now installs ffmpeg 5, which causes at least some tests to fail (notably the CLI test that predicts on an actual video). More work is needed to understand what differences there are in the new version that change the model output and whether it's worth supporting this updated version.

PR #400, opened 2026-02-16
Improved dependency handling

pjbull:
Lots of updates with the goals of:

 - Separate image/video extras and dependencies (along with tests for isolated envs)
 - Move to pixeltable-yolox, which is pip-installable and builds on Windows
 - Windows support (partly from above)
 - Make zamba pip-installable again

Changelog:

 - Split core dependencies into optional extras: `video` (av, ffmpeg-python, pytorchvideo, pixeltable-yolox, etc.), `image` (megadetector, Pillow), `tests` (pytest, black, flake8, coverage, nvidia-ml-py, etc.), and `docs` (mkdocs, mike, mkdocstrings). Base install no longer pulls in video/image stacks; use `pip install zamba[video]`, `zamba[image]`, or `zamba[video,image]`.
 - Remove `requirements-dev.txt` and `requirements-dev/`; use `pip install -e ".[tests]"` for development and `pip install -e ".[docs]"` for building docs.
 - Replace `mlflow` with `mlflow-skinny` in core dependencies. Add `pyarrow>=23.0.0`. Add Windows-specific torch version constraint for gloo bug. Declare Python 3.11–3.13 support and `requires-python = ">=3.11, <3.14"`.
 - Remove DensePose from optional dependencies so the package is publishable to PyPI. DensePose (detectron2, detectron2-densepose) must be installed from GitHub; see docs. Running `zamba densepose` without them prints install instructions and exits. Add `[tool.uv.extra-build-dependencies]` for detectron2 (torch) so CI can build it from GitHub.
 - Pin `setuptools<82` so the image extra works on Python 3.12 (megadetector’s yolov5 stack requires `pkg_resources`, which setuptools 82 removed). Add `[tool.uv]` override-dependencies for protobuf and setuptools.
 - Add `zamba.models.config_common` with shared types and validators (ModelEnum, MonitorEnum, ZambaBaseModel, RegionEnum, get_filepaths, validate_model_cache_dir, etc.). Move video-specific file/checkpoint logic into `zamba.models.config` and reuse common helpers.
 - Add `zamba.models.instantiation` and move `instantiate_model`, head-replacement, and resume logic out of `model_manager` for clearer separation.
 - Add lazy model registration in `zamba.models.registry` (`ensure_registered()`): video model classes are imported only when needed so the package can be imported without video dependencies. Config validation calls `ensure_registered()` when resolving checkpoint/model name.
 - Register image and utils sub-apps lazily so `zamba --help` and non-image/non-utils commands work without image or densepose dependencies. Defer imports of VideoLoaderConfig and config classes into the train, predict, and depth command callbacks.
 - Handle missing DensePose dependencies in the `densepose` command: catch ImportError when running the model and print install-from-GitHub instructions and exit with code 1.
 - Make NPY cache path hashing stable: use a JSON-serializable, order-invariant representation of the config (including Enum and Path) instead of `str(hashed_part)`. For local absolute paths, use a relative-style path in the cache key.
 - Fix cache cleanup in `npy_cache.__del__` by comparing `Path(cache_path).parents[0]` with `Path(tempfile.gettempdir())`.
 - Import image classifier lazily from `zamba.images` to avoid pulling in torch at package import time. Images config and manager use `zamba.models.config_common` and `zamba.models.instantiation` instead of config/utils from video.
 - Make megadetector import resilient: try `megadetector.detection.run_detector`, fall back to `detection.run_detector`. Make MLflow optional in image training (log warning and continue without it if import or setup fails). Disable `torch.compile` on Windows in addition to macOS.
 - Add pytest markers `video` and `image` and skip test files when the corresponding extra is not installed (conftest `collect_ignore` based on `_HAS_VIDEO` / `_HAS_IMAGE`). Define video-only fixtures and the dummy video model only when video deps are present. Set `CUDA_VISIBLE_DEVICES=0` in conftest to avoid DDP issues under pytest-xdist.
 - Add Makefile targets `tests-fast` (fail on first failure) and `test-image-only` / `test-video-only` (isolated venvs with only image or video extra). CI runs these isolation steps and installs image,video extras for the main test matrix. DensePose tests install detectron2 from GitHub with `--no-build-isolation` instead of using a densepose extra.
 - Update tests to use new config/instantiation imports and markers; add image dataset import check in the install smoke test.
 - Installation docs: document optional extras (video, image), PyPI install (`pip install zamba[video]` etc.), and Windows (image extra no extra tools; video needs Visual Studio Build Tools and FFmpeg). State FFmpeg only required for video workflows. Contribute page: dev install with `.[tests,image,video,docs]`, DensePose deps from GitHub, `make requirements` with uv. DensePose doc: install detectron2/detectron2-densepose from GitHub; note that `zamba densepose` prints instructions if deps missing.
 - Re-enable PyPI publish steps in the release workflow (Test PyPI and Production PyPI). Docs workflows and release job install `.[docs]` instead of `requirements-dev/docs.txt`.


pjbull:
@ejm714 This should be ready to go, I ran the test suite on a Windows machine, a Mac, and a linux box with a GPU. 

The diff looks big, but almost all of the substantive code changes are just refactoring so that image/video code paths are separable and imports are lazy loaded if need be so that you don't have to have all the deps for both sides available. Other code changes are almost all about stabilizing the test suite across platforms.

Easiest way to start the review may be to look at the pyproject.toml changes, the Makefile changes, and the README changes to see what surface area changed and then the things in the weeds will make more sense.

Don't want to rock the boat if you're demoing this week but could be nice to have the new version for folks pip-installable!

docs/docs/contribute/index.md
```diff
@@ -11,20 +11,29 @@ There you can find [open issues](https://github.com/drivendataorg/zamba/issues)
...
-To install `zamba` for development, you need to clone the git repository and then install the cloned version of the library for local development.
+To install `zamba` for development, clone the git repository and install the package with the dev extras. We recommend [uv](https://docs.astral.sh/uv/) for managing the environment.
 
-To install for development:
+With uv (recommended):
 ```console
 $ git clone https://github.com/drivendataorg/zamba.git
 $ cd zamba
-$ pip install -r requirements-dev.txt
+$ uv pip install -e ".[tests,image,video,docs]"
```
ejm714, line 20: maybe worth creating a "dev" dependency group for simpler install? https://docs.astral.sh/uv/concepts/projects/dependencies/#nesting-groups

docs/docs/install.md
```diff
@@ -20,46 +20,54 @@ We recommend [Python installation using Anaconda](https://www.anaconda.com/downl
  - [macOS installation video](https://www.youtube.com/watch?v=nVlrpNf3EdM)
 
 
-#### FFmpeg version 4
+#### FFmpeg > 4.3 (only needed for video workflows)
```
ejm714, line 23: I think we saw different predictions with ffmpeg 5 https://github.com/drivendataorg/zamba/issues/180 that were sufficient enough to change the top class. Probably needs more testing before we can remove the ceiling

tests/test_filter_frames.py
```diff
@@ -71,7 +74,7 @@ def test_weighted_euclidean(mdlite, frames, detections):
     assert (
         filtered_frames[:, 0, 0, 0]
         == np.array(
-            [90, 80, 70, 60, 44, 67, 73, 5, 54, 64, 65, 34, 93, 72, 56, 50, 87, 83, 47, 88]
+            [90, 80, 70, 60, 72, 96, 66, 75, 95, 93, 69, 23, 76, 57, 45, 83, 50, 51, 56, 73]
```
ejm714, line 77: are these differences concerning or expected? want to make sure we're not dramatically changing the output of MDLite where it's going to affect overall classification accuracy

tests/test_filter_frames.py
```diff
@@ -84,6 +87,6 @@ def test_weighted_prob(mdlite, frames, detections):
     assert (
         filtered_frames[:, 0, 0, 0]
         == np.array(
-            [90, 80, 70, 60, 50, 87, 30, 40, 34, 10, 22, 20, 71, 16, 39, 14, 77, 65, 42, 13]
+            [90, 80, 70, 60, 50, 13, 30, 40, 49, 10, 34, 14, 52, 36, 81, 99, 79, 24, 0, 20]
```
ejm714, line 90: same Q as above

zamba/models/densepose/densepose_manager.py
```diff
@@ -40,6 +40,19 @@
 from zamba.models.utils import RegionEnum, download_weights
 from zamba.settings import get_model_cache_dir
 
+# Message shown when DensePose deps are not installed (not on PyPI; install from GitHub).
+DENSEPOSE_INSTALL_MESSAGE = """
+DensePose dependencies are not installed. They are not distributed on PyPI.
+Install them from GitHub after installing PyTorch:
+
+  pip install torch
```
ejm714, line 48: worth having this be `uv pip` to match docs?

pyproject.toml
```diff
@@ -106,4 +121,14 @@ allow_redefinition = true
...
-addopts = "--cov-append --cov=zamba --cov-report=term --cov-report=html --cov-report=xml -n=auto --report-log reportlog.jsonl"
\ No newline at end of file
+addopts = "--cov-append --cov=zamba --cov-report=term --cov-report=html --cov-report=xml -n=auto --report-log reportlog.jsonl"
+markers = [
+  "video: tests that require the video extra (av, ffmpeg, opencv, yolox, pytorchvideo)",
+  "image: tests that require the image extra (megadetector, Pillow)",
+]
+
+[tool.uv]
+override-dependencies = [
```
ejm714, line 131: TWIL!

Makefile
```diff
@@ -69,9 +65,13 @@ lint:
 tests: clean-test
 	pytest tests -vv
 
+## Dev: fail fast on the first test failure
+tests-fast: clean-test
```
ejm714, line 69: [nit] should we standardize so all test commands are test (singular) not test? 

- make test
- make test-fast
- make test-debug
- make test-image-only
- make test-video-only

Makefile
```diff
@@ -81,6 +81,20 @@ images-tests: clean-test
 densepose-tests:
 	ZAMBA_RUN_DENSEPOSE_TESTS=1 pytest tests/test_densepose.py tests/test_cli.py::test_densepose_cli_options -vv
 
+## Test image deps in isolation: fresh venv, install image extra only, run image + agnostic tests, cleanup
+test-image-only:
```
ejm714, line 85: looks like this is duplicative of `images-tests` which i don't think we need any more. can you double check all test commands in the makefile?

HISTORY.md
```diff
@@ -1,5 +1,27 @@
...
+ - Make megadetector import resilient: try `megadetector.detection.run_detector`, fall back to `detection.run_detector`. Make MLflow optional in image training (log warning and continue without it if import or setup fails). Disable `torch.compile` on Windows in addition to macOS.
+ - Add pytest markers `video` and `image` and skip test files when the corresponding extra is not installed (conftest `collect_ignore` based on `_HAS_VIDEO` / `_HAS_IMAGE`). Define video-only fixtures and the dummy video model only when video deps are present. Set `CUDA_VISIBLE_DEVICES=0` in conftest to avoid DDP issues under pytest-xdist.
+ - Add Makefile targets `tests-fast` (fail on first failure) and `test-image-only` / `test-video-only` (isolated venvs with only image or video extra). CI runs these isolation steps and installs image,video extras for the main test matrix. DensePose tests install detectron2 from GitHub with `--no-build-isolation` instead of using a densepose extra.
```
ejm714, line 20: flagging that if we rename commands in the makefile, need to reflect that here too

docs/docs/install.md
```diff
@@ -1,6 +1,6 @@
 # Installing `zamba`
 
-Zamba has been developed and tested on macOS and Ubuntu Linux for both CPU and
+Zamba has been developed and tested on macOS, Ubuntu Linux, and Windows for both CPU and
```
ejm714, line 3: from a user-perspective, this file would probably be easier to follow if the departure point was the operating system rather than the install step.

so instead of 
```
step 1
- linux
- macos
- windows
```

it was

```
linux
- step 1
- step 2
- step 3

macos
- ...
```

ejm714 (changes requested):
So much good stuff in here! Just a couple small things

pjbull:
> are these differences concerning or expected? want to make sure we're not dramatically changing the output of MDLite where it's going to affect overall classification accuracy

@ejm714 I did some comprehensive testing to be sure. I identified that out of a 1000 video sample from a diversity of data sources, codecs, containers, file extensions, etc. there were ~100 that had _different_ predictions.

I dug into _why_ the predictions were different, and the root cause was actually a pandas version issue, not an ffmpeg issue. The default tie-break rules on sorting were different across pandas 2/3. That is now fixed to be deterministic [regardless of version](https://github.com/drivendataorg/zamba/pull/400/changes/f4958af69260bf75c8fd9782efc81068815efd62).

After that fix, we see the same top-1 for the 1000 test videos so I think we are good to go here after the tests pass

